PR #16, opened 2020-11-20
Outdated tutorial

kanatliemre:
I changed a few things from the top of my head, but might need some correction.


cgcardona (approved):
Looks good. :shipit:

PR #29, opened 2020-12-03
Updated Getting Started tutorial

jpop32:
-included information about choosing the proper architecture, building
vs using pre-built binaries, and staking using the wallet
-moved go installation instructions under building from source section
-removed no longer applicable parts

### DIFF
--- a/build/getting-started.md
+++ b/build/getting-started.md
@@ -16,7 +16,7 @@ If your issue isn’t addressed in the FAQ, come ask for help in the [Avalanche 
 Avalanche is an incredibly lightweight protocol, so the minimum computer requirements are quite modest.
 
 * Hardware: CPU &gt; 2 GHz, RAM &gt; 4 GB, Storage &gt; 10 GB free space
-* OS: Ubuntu 18.04/20.04 or Mac OS X &gt;= Catalina
+* OS: Ubuntu 18.04/20.04 or MacOS &gt;= Catalina
 
 ## Run an Avalanche Node and Send Funds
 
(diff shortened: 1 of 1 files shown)


build/getting-started.md
```diff
@@ -83,23 +87,23 @@ If you built from source:
 ./build/avalanchego
 ```
 
-If you are using the released binaries:
+If you are using the pre-built binaries on Mac OSX:
```
danlaine, line 90: Nit: Apple has rebranded to MacOS

jpop32: Tru dat.

build/getting-started.md
```diff
@@ -83,23 +87,23 @@ If you built from source:
...
 ```
 
-If you want to be able to make API calls to your node from other machines, include argument `--http-host=` \(e.g. `./build/avalanchego --http-host=`\)
+If you are using the pre-built binaries on Linux:
 
-You can use `Ctrl + C` to kill the node.
-
-To connect to the Fuji Testnet instead, use argument `--network-id=fuji`. You can get funds on the Testnet from the [faucet.](https://faucet.avax-test.network/)
+```cpp
+./avalanchego-<VERSION>-linux.4/avalanchego
```
danlaine, line 99: I think the `.4` is some kind of typo in the 1.0.6 release, so we'll probably remove that. Checking about that now

jpop32: I noticed that, but it's what's in the release archives right now. Also, the Mac version has files inside the build/ directory, Linux does not. It could use some standardization to simplify things.

I would leave it as it is now, not to confise people following the tutorial, and for future releases coordinate with the Ops to make each platform follow the same convention.

danlaine (approved):

PR #44, opened 2020-12-20
Update tutorials referencing C-Chain funding

jpop32:
- updated screenshots with the latest UX and flow
- updated the remaining tutorials that reference funding on the C-Chain to include link to Wallet
- also did a bit of formatting (added headings where missing and such)
- moved main net settings before test and local networks


cgcardona (approved):
Thanks Jovica!

PR #66, opened 2021-01-19
Update AVM Transaction Serialization docs with learnings from ongoing trustwallet work 

EvanJRichard:
## Summary
I've been using [the docs](https://docs.avax.network/build/references/avm-transaction-serialization) (along with our code) to figure out how to implement our notion of a Transaction for the x-chain for the [Trustwallet wallet-core work](https://github.com/EvanJRichard/wallet-core). While doing so, I noticed a bug that some fields that should be sorted were unsorted. I fixed that, and while I was at it I added some quality of life improvements and consistency tweaks that I hope will make it easier for the next person who comes along to work from these docs: 

### Changelog

- encoding is sometimes not correct, fix it
- we often to refer to Output, clarify as TransferableOutput vs TransactionOutput vs the Output that NFTMintOp uses
- definition of TransferableInput says "This will be a SECP256K1 etc." - clarify as "At time of writing this can only be a SECP256K1 etc."
- clarify when we mean TransferableInput vs Input, TransferableOp vs Operation, TransferableOutput vs Output
- NFTMintOp references an Output object that is very different from what we mean when we say Output elsewhere in the docs, highlight that this is the case
- Clarify that InitialState wants an Output not a TransferableOutput
- Sometimes the example for an object uses backtick code to define the variables, sometimes not. Examples standardized to one style.
- Add in-document markdown anchors to clarify things/provide cross-references

### Testing
No code, so no testing. Compiled the markdown locally and poked around, nothing obviously wrong or broken.

### See also
I didn't see any other encoding errors but I worry some still lurk.

### DIFF
--- a/build/references/avm-transaction-serialization.md
+++ b/build/references/avm-transaction-serialization.md
@@ -143,7 +143,7 @@ Let’s make a transferable input:
     // input:
     0x00, 0x00, 0x00, 0x05, 0x00, 0x00, 0x00, 0x00,
     0x07, 0x5b, 0xcd, 0x15, 0x00, 0x00, 0x00, 0x02,
-    0x00, 0x00, 0x00, 0x07, 0x00, 0x00, 0x00, 0x03
+    0x00, 0x00, 0x00, 0x03, 0x00, 0x00, 0x00, 0x07
 ]
 ```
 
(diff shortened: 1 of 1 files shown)


cgcardona (approved):
Great work. Thanks 🎩

PR #68, opened 2021-01-22
Deep links need to be lowercase.

cgcardona:


danlaine:
Why are we deleting the Platform VM info?

cgcardona:
We're deleting a duplicate and unused file. The actual file which we use is https://github.com/ava-labs/avalanche-docs/blob/master/build/references/platform-transaction-serialization.md

danlaine (approved):

PR #94, opened 2021-02-16
Fix the endpoint for the example call to eth_getAssetBalance on C-Chain

benber86:
The example call in the docs gives an endpoint on the X chain, so it will fail if someone just pastes the sample curl code


cgcardona (approved):
Nice catch!

PR #103, opened 2021-03-04
Japanese Translation

Avalanche-Rocky:
Please review before merging.


jpop32:
Looks good, but I'm having trouble parsing themore complex stuff just by looking at Markdown, will re-check once it's on GitBook.

PR #109, opened 2021-03-11
Update graph terminology in consensus docs

clydedacruz:
A block has one parent and one child, all subsequent blocks would be descendents.


danlaine (approved):

danlaine:
thanks @clydedacruz

PR #112, opened 2021-03-20
Emergency downgrade script + procedure

jzu:


cgcardona:
This looks good to me. @jpop32 thoughts?

jpop32:
Talked it over with Dan, and though this is a fine solution, it might be better to merge it into the existing installer script. So we'll use this approach and just add it to the other script.

cgcardona:
@jpop32 can you handle merging this into an existing script and creating a new PR?

jpop32:
Yep, will do.

jpop32:
Added downgrade functionality to the installer script #139.

Thanks @jzu for the inspiration. Also pilfered parts of your script. :-) Credited in the tutorial, of course.

PR #135, opened 2021-04-13
Update upgrade-your-avalanchego-node.md

tendon:
When upgrading from source at times the latest release is not always merged into master. Please add the proposed notes so that we can run the latest release from source, but by checking out the release tag commit rather than master.


tendon:
Just realized that the instructions I added should also have a `git fetch --all --tags` too, before the `git checkout...`

cgcardona (approved):
Thanks!

PR #141, opened 2021-04-25
Removed avash dependency

Devilla:
## Changelog 📝  

- [X]  Avash dependence removed. ❌  

Note: No need to use `./avash`, run an avalanche node documentation is present with `./avalanchego`. ✅


cgcardona:
Can you give some more context here? If we remove the steps related to avash and a person doesn't have prior knowledge about running an avalanchego instance then this tutorial will break.

Devilla:
> Can you give some more context here? If we remove the steps related to avash and a person doesn't have prior knowledge about running an avalanchego instance then this tutorial will break.

Sure.

cgcardona:
Checking in before I close this to inactivity.

Devilla:
> Checking in before I close this to inactivity.

okay, @cgcardona

PR #143, opened 2021-04-30
Update Ledger tutorial

jpop32:
Not that the Ledger app is out of experimental mode, tutrial needed to be udpated to reflect that.


cgcardona (approved):
Looks great TY!

PR #149, opened 2021-05-05
Update upgrade-your-avalanchego-node.md

tendon:
Added the `--force` flag to the `git checkout`. This will wipe out any local changes that you made (though probably you shouldn't have done any if this is a running node.) This is necessary because this morning when doing the v1.4.0 upgrade the checkout command informed that:

error: Your local changes to the following files would be overwritten by checkout:
	go.sum
Please commit your changes or stash them before you switch branches.
Aborting

I'm not familiar with golang, but understand that go.sum is about checksums of dependent modules. Was it modified or generated by running the build or the node? Would it make sense to either include go.sum in the committed version or to list it in the gitignore as to avoid that error message (in which case this pull request might not be needed.)

### DIFF
--- a/build/tutorials/nodes-and-staking/upgrade-your-avalanchego-node.md
+++ b/build/tutorials/nodes-and-staking/upgrade-your-avalanchego-node.md
@@ -148,7 +148,7 @@ Pull the latest code:
 git pull
 ```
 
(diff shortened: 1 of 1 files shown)


cgcardona (approved):
Thx 🏅

PR #168, opened 2021-05-24
Update doc for pre/post-1.4.5

yulin-dong:
Update the doc so that it applies to node upgrade from any version of pre-1.4.5 to that of post-1.4.5


cgcardona (approved):

cgcardona:
Thx! 🏅

PR #172, opened 2021-05-26
Doc: update db migration with step by step instruction

yulin-dong:
I moved the detailed instructions into the public doc.


cgcardona (approved):
🏆

PR #222, opened 2021-07-31
Fix initial empty network dashboard

quasarati:
Fix default time period to 1 hour relative  instead of fixed time period in the past to prevent an initial empty dashboard

### DIFF
--- a/dashboards/network.json
+++ b/dashboards/network.json
@@ -2618,8 +2618,8 @@
     "list": []
   },
   "time": {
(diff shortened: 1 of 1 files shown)


quasarati:
See discord chat here for a full description of the problem and the journey it took for a solution:
https://discord.com/channels/578992315641626624/620633143002660874/870971195816812614

danlaine (approved):

danlaine:
thanks @quasarati !

PR #308, opened 2021-09-10
docs: change uptime requirements from 60% to 80%

yulin-dong:
I did a search of 60% and replaced all places in that search result. Please let me know if there is any other places we need to change.


danlaine (approved):

danlaine:
thanks @yulin-dong

PR #337, opened 2021-09-25
Update avm-api-x-chain.md

glefait:
Remove `initialHolders` as it is not part of the `createNFTAsset` specification.
Fix some trivial typos.


danlaine:
merci @glefait

PR #339, opened 2021-09-26
update dashboards to include app messages

danlaine:


jpop32:
A nit: It's not entirely clear what 'app messages' are. That's VMs communicating with each other, in this case gossiping txs?

Also, dashboards should include refresh rates, 10s is ok. As it is, looks like we lost that on some of them so they don't refresh at all.

jpop32 (approved):
lgtm.

PR #349, opened 2021-10-04
Tutorials contest

jpop32:
Added the tutorials contest section with an intro page linking out to individual contributions in the subdirectories.

Linked tutorials are not referenced in the SUMMARY page so they won't render as GitBook pages, but as regular Markdown in GitHub interface, to make them distinct from the 'official' pages.


cgcardona (approved):
Good to go! 🚀

PR #380, opened 2021-11-04
zh-cn Lilt Translations PR

lilt-connector:


yulin-dong (changes requested):
Please don't use the same title for all your PRs, at least please provide info on what language this is about.

PR #401, opened 2021-11-21
Update monitoring-installer.sh. Improve idempotency of script

tdorzhi:
Add `-p` flag for mkdir even if 1-level directory created. This ensures script doesn't exit fail fast if directory already exists

Add check if `prometheus` user already exists. If not script won't try to create one and fail fast because it already exists

### DIFF
--- a/scripts/monitoring-installer.sh
+++ b/scripts/monitoring-installer.sh
@@ -84,11 +84,11 @@ install_prometheus() {
   fi
   echo "Attempting to download: $promFileName"
   wget -nv --show-progress -O prometheus.tar.gz $promFileName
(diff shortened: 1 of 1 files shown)


jpop32:
Requested changes are not problematic, but the script itself is not intended (nor tested) to be idempotent. As it is, I think it may be better to fail early if the app is already installed than to proceed and possibly mess up the already present installation.

For example, step 3 blindly adds `avalanchego` and `avalanchego-machine` jobs to prometheus config file. Running that twice will surely break prometheus.

If you would be willing to test the script and harden it against such situations, that would be great. Even just failing with a correct message would be an improvement.

PR #461, opened 2021-12-14
Granular api config

aaronbuchwald:
This is a PR to update the documentation for the C-Chain config file. We are adding more granular support for enabling specific services on the C-Chain instead of enabling entire namespaces. I added documentation specifying exactly which calls are added to which namespaces based on the config.

We could probably improve the style of this documentation and may want to consider providing links for each API call to either our own documentation or the relevant Ethereum API call documentation. Please leave any feedback you may have on the best way to organize this.

This PR is also blocked on actually merging the granular API config changes in coreth and bumping the AvalancheGo dependency to use the updated version.


yulin-dong (commented):
https://avalancheavax.slack.com/archives/C020PJ1U9CZ/p1639774154475600?thread_ts=1639773918.474100&cid=C020PJ1U9CZ

@Aaron Buchwald we’ll also need to update this to use the new array approach: https://github.com/ava-labs/avalanche-docs/pull/461/files

StephenButtolph:
1 nit: https://github.com/ava-labs/avalanche-docs/blob/granular-api-config/docs/build/tutorials/smart-contracts/using-truffle-with-the-avalanche-c-chain.md?plain=1#L259 should be updated as well I think

StephenButtolph (approved):

yulin-dong (approved):

PR #474, opened 2021-12-20
fix timestampvm api calls

Nuttymoon:
Both of the `timestampvm.getBlock` API snippets return an error in current version:
```sh
curl -X POST --data '{
    "jsonrpc": "2.0",
    "method": "timestampvm.getBlock",
    "params":{
        "id":""
    },
    "id": 1
}' -H 'content-type:application/json;' "127.0.0.1:9650/ext/bc/$BC_ID"
{"jsonrpc":"2.0","error":{"code":-32000,"message":"couldn't unmarshal an argument. Ensure arguments are valid and properly formatted. See documentation for example calls","data":null},"id":1}
```
I corrected the API calls by removing the `"id":""` param which shouldn't be specified when looking for the latest block.

Also moved a misplaced comma in another code block.


yulin-dong:
@ceyonur can you review this?

ceyonur (approved):
lgtm

yulin-dong (approved):

PR #516, opened 2022-01-14
Update run-avalanche-node for Go verision1.17.x

Domdu:
'go get' has been deprecated for installing binaries since Go 1.17 and will be become impossible in Go 1.18. For 1.17 'export GO111MODULE=off' is enough. But with go 1.18 'go get' command need to change with 'go install' command.


yulin-dong:
This is very old PR. Feel free to reopen if needed.

PR #518, opened 2022-01-17
Update create-a-virtual-machine-vm.md

Domdu:
SetStatus function missing in documentation.

### DIFF
--- a/docs/build/tutorials/platform/subnets/create-a-virtual-machine-vm.md
+++ b/docs/build/tutorials/platform/subnets/create-a-virtual-machine-vm.md
@@ -689,6 +689,9 @@ func (b *Block) Status() choices.Status { return b.status }
 
 // Bytes returns the byte repr. of this block
 func (b *Block) Bytes() []byte { return b.bytes }
+
+// SetStatus sets the status of this block
+func (b *Block) SetStatus(status choices.Status) { b.status = status }
 ```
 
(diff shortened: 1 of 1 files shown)


ceyonur:
set status is not a part of `snowman.Block` interface. It is a helper function, like `block.Initialize`. We should create a new subsection for utility functions. Thanks for spotting it.

ceyonur:
@Domdu feel free to re-open this or create a new PR with the new section for these functions.

Domdu:
Hi Ceyhun,

Correct me if I wrong, but ID and Status functions are not part of of snowman.Block interface but you are point these functions in this documentation block. I think this function must include this implementation because in Reject function you use SetState function. 

As a developer I want to follow your instructions and develop without any miss declarations. But some part of this document is missing like const variables or some part of this document not updated and mismatch with https://github.com/ava-labs/timestampvm. 

To sum up, when I follow these instructions my final result must be same with your sample project timestampvm. Without this imlementations I think this document will be missing.

ceyonur:
Hi,
ID and Status are a part of `choices.Decidable` which is a part of `snowman.Block`, so transitively they're a part of `snowman.Block`.

Thank you for pointing it, we will update the document.

Domdu:
I deleted my comment after realize choices.Decidable interface. Thanks for response :)

Issue #462, opened 2021-12-14
Using Truffle with the Avalanche C-Chain not working


israel-gonzalezmedina:
I have a set of smart contracts that work very well with ganache. Now I am trying to run them with avalanche and I am following the documentation in:
https://docs.avax.network/build/tutorials/smart-contracts/using-truffle-with-the-avalanche-c-chain
I am using Avast like written in the doc, even in other places it says Avast is deprecated, however, the documentation with truffle is set for avast. When I try to use the network with the command:

          npx truffle console --network development

I get the error:
```
Error: ProviderSubprovider - specified provider does not have a sendAsync method
    at new ProviderSubprovider (/home/user/node_modules/web3-provider-engine/subproviders/provider.js:12:34)
    at new HDWalletProvider (/home/user/node_modules/@truffle/hdwallet-provider/src/index.ts:255:31)
    at Object.provider (/home/user/truffle-config.js:69:16)
    at Object.getProvider (/home/user/node_modules/truffle/build/webpack:/packages/provider/index.js:20:1)
    at Object.create (/home/user/node_modules/truffle/build/webpack:/packages/provider/index.js:13:1)
    at TruffleConfig.get [as provider] (/home/user/node_modules/truffle/build/webpack:/packages/config/dist/configDefaults.js:213:1)
    at Object.detect (/home/user/node_modules/truffle/build/webpack:/packages/environment/environment.js:19:1)
    at Object.run (/home/user/node_modules/truffle/build/webpack:/packages/core/lib/commands/console.js:47:1)
    at Command.run (/home/user/node_modules/truffle/build/webpack:/packages/core/lib/command.js:172:1)
    at Object.586806 (/home/user/node_modules/truffle/build/webpack:/packages/core/cli.js:57:1)
    at __webpack_require__ (/home/user/node_modules/truffle/build/webpack:/webpack/bootstrap:18:1)
    at __webpack_require__.x (/home/user/node_modules/truffle/build/webpack:/webpack/bootstrap:36:1)
    at Function.__webpack_require__.x (/home/user/node_modules/truffle/build/webpack:/webpack/runtime/startup chunk dependencies:36:1)
    at /home/user/node_modules/truffle/build/webpack:/webpack/startup:3:1
    at Object.<anonymous> (/home/user/node_modules/truffle/build/cli.bundled.js:640:12)
    at Module._compile (internal/modules/cjs/loader.js:701:30)
[...]

PR #522, opened 2022-01-17
#462 - update truffle config to work with default installed Web3 (#1)

michaelBenin:
* #462 - update truffle config to work with default installed Web3

* ava-labs#462 - promisify sendAsync method for truffle


yulin-dong:
We have re-organize our doc site. If you still want this to be considered, please create a new PR against our new project structure.

PR #528, opened 2022-01-19
Update create-a-virtual-machine-vm.md

Domdu:
Connected function need nodeVersion as a parameter.


ceyonur:
this is currently not true until we bump the avalanchego version in the timestamp vm. Plus we explicitly mention that the docs uses a specific version of timestampvm and avalanchego. so we should also change that line.

https://github.com/ava-labs/avalanche-docs/blob/88e47ce402cc7b611a37e820ae6a96a0ffbd4d53/docs/build/tutorials/platform/subnets/create-a-virtual-machine-vm.md?plain=1#L22

ceyonur:
Closing in favor of #530

PR #590, opened 2022-02-17
change file name command-line-interface to avalanchego-config-flags

yulin-dong:


yulin-dong:
> Other than changing `config flag argument` --> `config flag` LGTM

Fixed

PR #631, opened 2022-03-14
Add note on how VMID is calculated

aaronbuchwald:
This PR adds brief documentation on how the VMID is calculated. We should add this to the FAQ as well, but after looking at the FAQ I think this should be part of a separate PR after we have separated out the subnet documentation from the subnet-evm documentation.


yulin-dong (approved):

rajranjan0608:
Really needed this explanation. Thanks 👍 .

rajranjan0608 (approved):

PR #671, opened 2022-04-05
Extracted chain configs into a separate document

jpop32:
- also rearranged the order so node and chain flags are at the top, as
they are by far the most popular reference files


docs/build/references/chain-config-flags.md
```diff
@@ -0,0 +1,448 @@
+---
+sidebar_position: 2
+description: Reference for all available chain config options and flags.
+---
+# Chain Configs
+
+Some chains allow the node operator to provide a custom configuration. AvalancheGo can read chain configurations from files and pass them to the corresponding chains on initialization.
+
+AvalancheGo looks for these files in the directory specified by `--chain-config-dir` AvalancheGo flag, as documented [here](avalanchego-config-flags.md#--chain-config-dir-string). If omitted, value defaults to `$HOME/.avalanchego/configs/chains`. This directory can have sub-directories whose names are chain IDs or chain aliases. Each sub-directory contains the configuration for the chain specified in the directory name. Each sub-directory should contain a file named `config`, whose value is passed in when the corresponding chain is initialized (see below for extension). For example, config for the C-Chain should be at: `{chain-config-dir}/C/config.json`.
```
yulin-dong, line 9:
```suggestion
AvalancheGo looks for these files in the directory specified by `--chain-config-dir` AvalancheGo flag, as documented [here](avalanchego-config-flags.md#--chain-config-dir-string). If omitted, value defaults to `$HOME/.avalanchego/configs/chains`. This directory can have sub-directories whose names are chain IDs or chain aliases. Each sub-directory contains the configuration for the chain specified in the directory name. Each sub-directory should contain a file named `config`, whose value is passed in when the corresponding chain is initialized (see below for extension). For example, config for the C-Chain should be at: `{chain-config-dir}/C/config.json`.

This also applies to subnets, for example, if a subnet's chain id is `2ebCneblahblahblah`, the config for this chain should be at `{chain-config-dir}/2ebCneblahblahblah/config.json`
```

yulin-dong (changes requested):

docs/build/references/chain-config-flags.md
yulin-dong: Should a new section/header for Subnet be created, as the same level of `C-Chain` and `X-Chain`?

yulin-dong (approved):

Issue #673, opened 2022-04-08
"http-host": "" by default is dangerous in avalanchego-installer.sh


jzu:
The `avalanchego-installer.sh` is a heavenly gift, all right. However, it creates a `.avalanchego/configs/node.json` config file which includes a `"http-host": ""` directive. This opens port 9650 on the Internet. I know validators should protect their servers using any flavor of iptables to weed out unwanted requests. It's not always the case, however. It happened to me (messed up my setup), it happened to someone else on Discord today (wasn't even aware of the situation). This is dangerous for the network, and I suggest removing this line from the script.

Consequences: logs show a bunch of

`Apr 02 03:53:23 avalanche avalanchego[1914078]: INFO [04-02|03:53:23.300] Submitted transaction                    hash=0xdb8ef72c29e74edd7a7a5548e0b01c127c85809e842b322e93a5114068679e21 from=0x000010ED59bCb8FAC3AECb568aefBf8b49daEA02 nonce=121 recipient=0x000010ED59bCb8FAC3AECb568aefBf8b49daEA02 value=0 type=2 gasFeeCap=63,363,090,303  gasTipCap=63,363,090,303  gasPrice=63,363,090,303`

then a bunch of

`Apr 02 04:22:42 avalanche avalanchego[1914078]: WARN [04-02|04:22:42.967] Served eth_call                          conn=103.100.174.14:61740 reqid=924 execTime="948.173µs" procTime="949.241µs" totalTime="949.722µs" err="execution reverted: INSUFFICIENT_OUTPUT" errdata=0x08c379a000000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000000000000000000000000000000000013494e53554646494349454e545f4f555450555400000000000000000000000000`

`health.health` give `"healthy": false`, node is benched. Restart the service, but `false` again is a matter of hours at most, rinse, repeat, until the directive is deleted from the config file or the validator sets up a correct iptables configuration.

jpop32:
This issue has been addresses with the new version of the installer script, it now prompts user for the RPC port binding config.

PR #674, opened 2022-04-09
Installer: remove http-host directive from config file (#673)

jzu:
See issue #673:  "http-host": "" by default is dangerous in avalanchego-installer.sh

See also [this Discord discussion on April 8](https://discord.com/channels/578992315641626624/620633143002660874/962019210790330448): this directive leads to validators leaving their node unprotected, unbeknownst to them.

Tested on an unsuspecting clean server.


jpop32:
There is a new version of the installer script in development, it will address this issue as well.

PR #707, opened 2022-04-24
DS-153: restructure existing doc contents by grouping them using top navigation bar

yulin-dong:


jpop32:
Yes, this is a great first step in streamlining our docs. Let's get this out and then we can get to reshaping the material into these categories: what goes where, what to merge, what to remove and what new material we need to create.

PR #730, opened 2022-05-07
fix chain config links

ceyonur:
fixes chain config links


yulin-dong (approved):

yulin-dong:
Thanks for catching this.

PR #824, opened 2022-06-07
Merge into ds 195

Julian-dev28:
Make 

Add a validator with AJS Doc even with master

> @Julian-dev28 please use master branch as base before requesting reviews

Sorry, It looks like it is automatic! did not tag for review


yulin-dong:
@Julian-dev28 please use master branch as base before requesting reviews

PR #869, opened 2022-06-27
add a warning notice for sharing the same node id

yulin-dong:


docs/nodes/maintain/node-backup-and-restore.md
```diff
@@ -16,6 +16,12 @@ The installation itself can be easily recreated by installing the node on a new
 
 ## NodeID
 
+:::warning
+
+If more than one running **validators** share the same NodeID, it will dramatically impact the uptime calculation of the validator which will very likely disqualify the validator from receiving the staking rewards. Please make sure only one validator with the same NodeID run at one time.
```
jpop32, line 21: Same goes for any node, not just validators. As I understood it, one of the validators will get ignored, but you cannot be sure which one it is. In any case, the gist of it is correct - if you do that you will most likely lose out on staking reward.

aaronbuchwald: I'd suggest also saying that the network will non-deterministically choose to connect to only one node with the same nodeID, but may connect/disconnect resulting in reduced uptime.

Also agree with @jpop32 that we should state this for any node as opposed to just validators.

yulin-dong: @aaronbuchwald can you review one more time, I updated it.

jpop32 (approved):
LGTM

PR #871, opened 2022-06-28
BlobVM

rajranjan0608:


docs/subnets/create-a-vm-blobvm.md
```diff
@@ -0,0 +1,454 @@
+# Create a Blob Virtual Machine (VM)
+
+[toc]
```
yulin-dong, line 3: This is not needed.

docs/subnets/create-a-vm-blobvm.md
```diff
@@ -0,0 +1,454 @@
...
+* Represent the operations in that state
+* Apply the operations in that state
+
+## BlobVM
+
+In this tutorial, we will learn all about virtual machines and how we can build them by taking references from **[BlobVM](https://github.com/ava-labs/blobvm)**. It's a virtual machine that can be used to instantiate key-value blockchains for storing files like images, videos, etc. in an efficient way.
+
+Blobs are small chunks of data. In BlobVM, we divide a file into small blobs and store them as a key-value pair. The key of these small chunks of a file is linked together as the children of the root. This tree is also stored as JSON data against the root's key.
+
+## How to use a VM
```
yulin-dong, line 20: Title case for all titles, please

docs/subnets/create-a-vm-blobvm.md
```diff
@@ -0,0 +1,454 @@
...
+	errs := svc.vm.Submit(tx)
+	if len(errs) == 0 {
+		return nil
+	}
+	if len(errs) == 1 {
+		return errs[0]
+	}
+	return fmt.Errorf("%v", errs)
+}
+```
```
cgcardona, line 330: [Fenced code blocks should be surrounded by blank lines](https://github.com/DavidAnson/markdownlint/blob/v0.25.1/doc/Rules.md#md031)

cgcardona (changes requested):

docs/subnets/create-a-vm-blobvm.md
```diff
@@ -0,0 +1,454 @@
...
+
+## BlobVM
+
+In this tutorial, we will learn all about virtual machines and how we can build them by taking references from **[BlobVM](https://github.com/ava-labs/blobvm)**. It's a virtual machine that can be used to instantiate key-value blockchains for storing files like images, videos, etc. in an efficient way.
+
+Blobs are small chunks of data. In BlobVM, we divide a file into small blobs and store them as a key-value pair. The key of these small chunks of a file is linked together as the children of the root. This tree is also stored as JSON data against the root's key.
+
+## How to use a VM
+
+VM modules are served as a binary plugin to the `avalanchego` software. The nodes that want to power their chains using a particular VM, must have the built binary in their avalanchego's `build/plugins` directory. There could be multiple VM plugins in this directory.
```
joshua-kim, line 22: This is dependent on the user's configuration. We can point them to the docs ([ref])(https://docs.avax.network/nodes/maintain/avalanchego-config-flags#build-directory), but this will default to `$GOPATH/src/github.com/ava-labs/avalanchego/build/plugins/` (if they haven't built from source), and `$SOURCE_DIR/build/plugins/` if they are building from source.

docs/subnets/create-a-vm-blobvm.md
```diff
@@ -0,0 +1,454 @@
...
+
+## High-Level Overview
+
+Virtual Machines are created as a module, whose binary is registered (loaded) by a node running `avalanchego`, against a user-defined **vmID** (binary file name must be vmID). We can build multiple chains with the registered VM. A VM is registered, only when its built binary is present in the `avalanchego/build/plugins/` directory.
```
joshua-kim, line 26: 1. nit: Seems confusing to word it as "registered (loaded)", prefer just saying "regsitered" or "loaded".
2. VM IDs are generated, not user-provided.
3. Same plugin directory comment as above.

docs/subnets/create-a-vm-blobvm.md
```diff
@@ -0,0 +1,454 @@
...
+* **Client** for making calls to the service's APIs
+
+### What are Handlers
+
+Handlers serve the response for the incoming HTTP requests. Handlers can also be wrapped with **gRPC** for efficiently making calls from other services such as avalanchego. Handler helps in creating APIs. VM implements 2 kinds of handlers -
+
+* **Non-Static Handlers** - They help in interacting with blockchains instantiated by the VM. The API's endpoint will be different for different chains. `/ext/bc/[chainID]`
+* **Static Handlers** - They help in directly accessing VM. These are optional for reasons such as parsing genesis bytes required to instantiate new blockchains. `/ext/vm/[vmID]`
+
+### Registering a VM
```
joshua-kim, line 40: I thought VM registry was something the user didn't care about (handled by avalanchego under-the-hood). Should this be in this tutorial? I don't see this in the `blobvm` code.

docs/subnets/create-a-vm-blobvm.md
```diff
@@ -0,0 +1,454 @@
...
+### Instantiating a VM
+
+Each VM has a **factory** that is capable of creating new VM instances. A VM can be initialized as a chain along with handlers for accessing it, only when we have the VM's instance. The factory's `New` method shown below, returns the VM's instance to its caller in `avalanchego`.
+
+```go
+// Returning a new VM instance from VM's factory
+func (f *Factory) New(*snow.Context) (interface{}, error) { return &vm.VM{}, nil }
+```
+
+VM's methods and other parts of the code are made available to `avalanchego` by **registering** its factory and instantiating a new VM using this factory whenever a chain is required to be built. It's generally in the [`factory.go`](https://github.com/ava-labs/blobvm/blob/master/factory.go) file of the VM.
```
joshua-kim, line 62:
```suggestion
```go
// Returning a new VM instance from VM's factory
func (f *Factory) New(*snow.Context) (interface{}, error) { return &vm.VM{}, nil }
```
[Reference](https://github.com/ava-labs/blobvm/blob/master/factory.go)

A VM's functionality is exposed to `avalanchego` by registering its factory and instantiating a new VM using this factory whenever a chain is required to be built.
```

File: docs/subnets/create-a-vm-blobvm.md [hunk 8]
```diff
@@ -0,0 +1,454 @@
...
+
+```go
+// Instantiating a new VM from its factory in avalanchego
+func (m *manager) buildChain(chainParams ChainParameters, sb Subnet) (*chain, error) {
+    ...
+    vmFactory, err := m.VMManager.GetFactory(vmID)
+    vm, err := vmFactory.New(ctx.Context)
+    ...
+}
```
joshua-kim, line 71: Let's link references to all code so it's easy to see where the example came from.

docs/subnets/create-a-vm-blobvm.md
```diff
@@ -0,0 +1,454 @@
...
+```
+
+### Initializing a VM
+
+Chains are functional, only when the instantiated VMs are initialized. Initializing a VM involves setting up the database, block builder, mempool, genesis state, handlers, etc. This will expose the VM's API handlers, start accepting transactions, gossiping them across the network, building blocks, etc.
```
joshua-kim, line 76: 1. I would prefer to not use "etc" since it implies that the user has some pre-existing knowledge of how VM's work. If they're reading the tutorial, we shouldn't make that assumption.
2. I think it would be more helpful to just explain what each parameter in `VM#Initialize` does/is used for.

docs/subnets/create-a-vm-blobvm.md
```diff
@@ -0,0 +1,454 @@
...
+    genesisData,
+    chainConfig.Upgrade,
+    chainConfig.Config,
+    msgChan,
+    fxs,
+    sender,
+);
+```
+
+You can refer to the implementation of `vm.initialize` in the [BlobVM](https://github.com/ava-labs/blobvm/blob/1fbb655246a2a41861f40d603ad40299b177252a/vm/vm.go#L92) repository.
```
joshua-kim, line 91: nit: Unsure if we should use a specific commit version as opposed to master@latest. Both have their pros/cons.
- latest's code is always up-to-date, but the docs will get stale as the vm is updated
- a commit's code is stale, but the docs will always be up-to-date

docs/subnets/create-a-vm-blobvm.md
```diff
@@ -0,0 +1,454 @@
...
+* **Client** - It is used for interacting with service APIs.
+* **Factory** - For creating new instances of the virtual machine.
+
+## Transaction Lifecycle
+
+Virtual Machine exposes APIs or handlers for users to make direct RPC to service or use the client to interact with the service. Every change on a chain can happen only through **transactions**. Let's see, how a transaction goes through the network to update the chain's state.
+
+* User calls `client.IssueRawTx` or directly make RPC to `service`
+* `service.IssueRawTx()` is called using handlers
+    * Receive Tx bytes as arguments
```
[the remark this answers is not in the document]
joshua-kim: +1

docs/subnets/create-a-vm-blobvm.md
```diff
@@ -0,0 +1,454 @@
...
+* **Block Builder** - Initiates the gossip of new transactions, notifies the engine that new blocks are ready to be built, and also handles next block generation notification.
+* **Storage** - To store and retrieve data of the chain's state.
+* **Virtual Machine** - Entry point for all the components to orchestrate them according to the consensus engine and API requests.
+* **Service** - API handlers for interacting with VM and initialized chain.
+* **Client** - It is used for interacting with service APIs.
+* **Factory** - For creating new instances of the virtual machine.
+
+## Transaction Lifecycle
+
+Virtual Machine exposes APIs or handlers for users to make direct RPC to service or use the client to interact with the service. Every change on a chain can happen only through **transactions**. Let's see, how a transaction goes through the network to update the chain's state.
```
joshua-kim, line 117: Not sure why we bold "transactions". We don't always bold variants of the word (ex. transaction later in the paragraph), so I think we can just not bold it entirely.

docs/subnets/create-a-vm-blobvm.md
```diff
@@ -0,0 +1,454 @@
...
+* **Block** - It handles block creation, initialization, verification, parsing, etc.
+* **Builder** - Returns the built preferred block by including transactions from the mempool.
+* **Block Builder** - Initiates the gossip of new transactions, notifies the engine that new blocks are ready to be built, and also handles next block generation notification.
+* **Storage** - To store and retrieve data of the chain's state.
+* **Virtual Machine** - Entry point for all the components to orchestrate them according to the consensus engine and API requests.
+* **Service** - API handlers for interacting with VM and initialized chain.
+* **Client** - It is used for interacting with service APIs.
+* **Factory** - For creating new instances of the virtual machine.
+
+## Transaction Lifecycle
```
joshua-kim, line 115: I'm a bit skeptical on if we need to go so deep in this tutorial on how avalanche handles your transactions under-the-hood. I feel like user's should just be concerned with how to implement a VM + transaction types and how avalanche handles it should just be a black-box that they don't care about.

docs/subnets/create-a-vm-blobvm.md
```diff
@@ -0,0 +1,454 @@
...
+* **Mempool** - Heap (min and max) for keeping pending transactions locally on the node.
+* **Network** - Implements gossip of transactions from mempool to the network.
+* **Block** - It handles block creation, initialization, verification, parsing, etc.
+* **Builder** - Returns the built preferred block by including transactions from the mempool.
+* **Block Builder** - Initiates the gossip of new transactions, notifies the engine that new blocks are ready to be built, and also handles next block generation notification.
+* **Storage** - To store and retrieve data of the chain's state.
+* **Virtual Machine** - Entry point for all the components to orchestrate them according to the consensus engine and API requests.
+* **Service** - API handlers for interacting with VM and initialized chain.
+* **Client** - It is used for interacting with service APIs.
+* **Factory** - For creating new instances of the virtual machine.
```
joshua-kim, line 113: IMO we should link specific interfaces that matter here, users might not know how to implement a Transaction/et al.

joshua-kim (changes requested):

docs/subnets/create-a-vm-blobvm.md
rajranjan0608: 2. Cool. Will add more explanations to it. VM IDs are generated from user-provided strings, right?

docs/subnets/create-a-vm-blobvm.md
rajranjan0608: Isn't transaction lifecycle defined by a VM? I have used VM components to explain how the transaction is processed within.

docs/subnets/create-a-vm-blobvm.md
yulin-dong: > 2. Cool. Will add more explanations to it. VM IDs are generated from user-provided strings, right?

yes, see https://docs.avax.network/subnets/subnet-cli#subnet-cli-create-vmid

docs/subnets/create-a-vm-blobvm.md
```diff
@@ -0,0 +1,470 @@
...
+);
+```
+
+You can refer to the implementation of `vm.initialize` in the [BlobVM](https://github.com/ava-labs/blobvm/blob/1fbb655246a2a41861f40d603ad40299b177252a/vm/vm.go#L92) repository.
+
+### Interaction
+
+Once a VM is initialized, it becomes a functional chain. We can use the handlers to issue transactions, query chain state, and whatever functionalities are being provided by the VM.
+
+### Consensus Engine
```
yulin-dong, line 97: This is just a header with no contents. Do you intent to add contents under it?

docs/subnets/create-a-vm-blobvm.md
```diff
@@ -0,0 +1,470 @@
...
+
+## Components of a Virtual Machine
+
+We have to implement the following components in our virtual machine.
```
yulin-dong, line 101: what does `our virtual machine` refer to?

docs/subnets/create-a-vm-blobvm.md
```diff
@@ -0,0 +1,470 @@
...
+- **Client** - It is used for interacting with service APIs.
+- **Factory** - For creating new instances of the virtual machine.
+
+## Transaction Lifecycle
+
+Virtual Machine exposes APIs or handlers for users to make direct RPC to service or use the client to interact with the service. Every change on a chain can happen only through **transactions**. Let's see, how a transaction goes through the network to update the chain's state.
+
+- User calls `client.IssueRawTx` or directly make RPC to `service`
+- `service.IssueRawTx()` is called using handlers
+  - Receive Tx bytes as arguments
```
yulin-dong, line 121: Should Tx be extended to `transaction` in text?

docs/subnets/create-a-vm-blobvm.md
yulin-dong: One suggestion: put what users must implement on the top, and use appendix for detailed explanation at the end of the doc.

docs/subnets/create-a-vm-blobvm.md
rajranjan0608: The VM we want to build.

yulin-dong: Then I would suggest to change to the following
```suggestion
We have to implement the following components in a new virtual machine.
```

rajranjan0608: Nice. Thanks

docs/subnets/create-a-vm-blobvm.md
```diff
@@ -0,0 +1,470 @@
...
+
+## High-Level Overview
+
+Virtual Machines are created as a module, whose binary is registered (loaded) by a node running `avalanchego`, against a user-defined **vmID** (binary file name must be vmID). We can build multiple chains with the registered VM. A VM is registered, only when its built binary is present in the `build/plugins/` directory.
+
+The VM exposes 3 things for interaction:
+
+- **Static and Non-Static Handlers (API)** for making requests to VM methods
+- **Service** for implementing handlers or APIs
+- **Client** for making calls to the service's APIs
```
aaronbuchwald, line 31: 3 things for interaction is strange wording to me, it sounds like this is the only purpose of the VM, but it doesn't include the actual logic for the consensus engine.

Can you re-phrase this to make it specific to the APIs exposed by the VM?

Also, the Client is not actually part of the VM interface, we have just implemented a client in the VM packages for calling the same APIs. I think it makes sense to include it in this tutorial, but we should clarify that it's not part of the VM itself.

aaronbuchwald: I think this piece could be under a section such as `Implementing APIs for BlobVM`

docs/subnets/create-a-vm-blobvm.md
aaronbuchwald: I think the chain manager itself may be better suited to a separate document or at the very least to an appendix/bottom of the doc as @yulin-dong  suggested.

docs/subnets/create-a-vm-blobvm.md
```diff
@@ -0,0 +1,470 @@
...
+    ...
+    vmFactory, err := m.VMManager.GetFactory(vmID)
+    vm, err := vmFactory.New(ctx.Context)
+    ...
+}
+```
+
+### Initializing a VM
+
+Chains are functional, only when the instantiated VMs are initialized. Initializing a VM involves setting up the database, block builder, mempool, genesis state, handlers, etc. This will expose the VM's API handlers, start accepting transactions, gossiping them across the network, building blocks, etc.
```
aaronbuchwald, line 76: Agree with @joshua-kim . I think a high level explainer of the tasks that a VM needs to fulfill and then subsections for each of those things that get straight to what the parameters are and what it needs to do would be good.

docs/subnets/create-a-vm-blobvm.md
```diff
@@ -0,0 +1,470 @@
...
+    fxs,
+    sender,
+);
+```
+
+You can refer to the implementation of `vm.initialize` in the [BlobVM](https://github.com/ava-labs/blobvm/blob/master/vm/vm.go#L92) repository.
+
+### Interaction
+
+Once a VM is initialized, it becomes a functional chain. We can use the handlers to issue transactions, query chain state, and whatever functionalities are being provided by the VM.
```
aaronbuchwald, line 95: This phrase once a VM is initialized it becomes a functional chain is repeated a few places, but it seems pretty vague to me. Additionally, the VM is initialized before being registered to the chain manager and isn't that functional until bootstrapping is finished, so I think it would be better to be more specific here.

docs/subnets/create-a-vm-blobvm.md
aaronbuchwald: Transactions are internal to the VM, the consensus engine only cares about blocks, so it's not necessary to go through it. The phrasing in this document suggests transactions are a hard requirement, so we should make sure that the flexibility of the VM interface is clear (or at least that we don't say that it's more rigid than it is).

docs/subnets/create-a-vm-blobvm.md
```diff
@@ -0,0 +1,854 @@
...
+	k := PrefixBalanceKey(address)
+	b := make([]byte, 8)
+	binary.BigEndian.PutUint64(b, bal)
+	return db.Put(k, b)
+}
+```
+
+- **ModifyBalance** - This function will modify the balance of an address depending upon the parameters `add` and `change`. It will simply perform the `SafeAdd` and `SafeSub` operation on the existing balance. `add` bool parameter indicates whether to add or to subtract the `change` amount from the existing balance. Finally it will call the `SetBalance` function with the updated amount.
+
+    It is called while executing the transactions. For transfer transactions it will [reduce](https://github.com/ava-labs/blobvm/blob/master/chain/transfer_tx.go#L39) the sender's balance and [increase](https://github.com/ava-labs/blobvm/blob/master/chain/transfer_tx.go#L42) the receiver's balance. And for every transaction it will be called to [reduce](https://github.com/ava-labs/blobvm/blob/master/chain/tx.go#L104) the fee from the sender.
```
cgcardona, line 501: Remove the spaces at the beginning of this paragraph.

docs/subnets/create-a-vm-blobvm.md
```diff
@@ -0,0 +1,854 @@
...
+
+```go
+vm.mempool = mempool.New(vm.genesis, vm.config.MempoolSize)
+```
+
+Whenever a transaction is submitted to VM, it first gets initialized, verified, and executed locally. If the transaction looks valid, then it is [added](https://github.com/ava-labs/blobvm/blob/master/vm/vm.go#L431) to mempool.
+
+```go
+vm.mempool.Add(tx)
+```
```
cgcardona, line 356: MD031/blanks-around-fences: Fenced code blocks should be surrounded by blank linesmarkdownlint[MD031](https://github.com/DavidAnson/markdownlint/blob/v0.25.1/doc/Rules.md#md031)

docs/subnets/create-a-vm-blobvm.md
```diff
@@ -0,0 +1,854 @@
...
+	balancePrefix = 0x4
+
+	linkedTxLRUSize = 512
+
+	ByteDelimiter byte = '/'
+)
+```
+
+We have 5 types of key-value pairs in total -
+* **Block** - For storing block data.
```
cgcardona, line 459: MD004/ul-style: Unordered list style [Expected: dash; Actual: asterisk]markdownlint[MD004](https://github.com/DavidAnson/markdownlint/blob/v0.25.1/doc/Rules.md#md004)

MD032/blanks-around-lists: Lists should be surrounded by blank linesmarkdownlint[MD032](https://github.com/DavidAnson/markdownlint/blob/v0.25.1/doc/Rules.md#md032)

cgcardona (changes requested):
Left a few more comments for small changes related to formatting markdown docs.

I suggest installing [markdownlint](https://marketplace.visualstudio.com/items?itemName=DavidAnson.vscode-markdownlint) as a VSCode plugin for linting markdown docs.

docs/subnets/create-a-vm-blobvm.md
rajranjan0608: Added more details at the bottom.

docs/subnets/create-a-vm-blobvm.md
rajranjan0608: Using master branch.

docs/subnets/create-a-vm-blobvm.md
rajranjan0608: Got it. Made it clear in the docs.

docs/subnets/create-a-vm-blobvm.md
```diff
@@ -0,0 +1,1672 @@
+# Create a Blob Virtual Machine (VM)
```
joshua-kim, line 1: High-level: 

1. I think we're going into too much detail in some parts of this doc. We should make a clear distinction between what must be implemented as part of a VM (e.g the VM's public interface), vs what was optionally added into the BlobVM and why.
2. We should run this through a grammar checker to get rid of awkward phrasing/run-on sentences/etc. I've used Grammarly and it's pretty nice. There might be some VS-Code plugin for it (I don't use VS code but I think the majority of the team does. Saw [znck](https://marketplace.visualstudio.com/items?itemName=znck.grammarly)  and [LTeX](https://marketplace.visualstudio.com/items?itemName=valentjn.vscode-ltex) looked like good candidates on google). 
3. Should we avoid duplicating common stuff in our [vm doc](https://docs.avax.network/subnets/create-a-virtual-machine-vm#introduction) and only put blob-vm specific stuff here? We can probably safely assume the reader has read that doc as a prereq, or post a link at the top of this doc to let the reader know they should be familiar with that doc before reading this.

docs/subnets/create-a-vm-blobvm.md
```diff
@@ -0,0 +1,1672 @@
...
+
+## Introduction
+
+Virtual Machine is a blueprint for a blockchain and a blockchain instantiates from VM, very much similar to a class-object relationship. Issuance of transactions, transaction types, block structure, block building algorithm, keeping pending transactions in the mempool, gossiping transactions to the connected nodes, etc. are defined in a virtual machine.
+
+Virtual Machines can be broken down into components like **Transaction**, **Mempool**, **Network**, **Block**, etc. Each component works together to:
```
joshua-kim, line 7: Avoid using "etc", since this is a tutorial we shouldn't presume knowledge from the reader. Also I don't think a VM's interface requires us to implement a mempool (even though it is a common abstraction used).

docs/subnets/create-a-vm-blobvm.md
```diff
@@ -0,0 +1,1672 @@
...
+
+- Define the representation of the blockchain's state
+- Represent the operations in that state
+- Apply the operations in that state
+
+## BlobVM
+
+In this tutorial, we will learn all about virtual machines and how we can build them by taking references from [BlobVM](https://github.com/ava-labs/blobvm). It's a virtual machine that can be used to instantiate key-value blockchains for storing files like images, videos, etc. in an efficient way.
+
+Blobs are small chunks of data. In BlobVM, we divide a file into small blobs and store them as a key-value pair. The key of these small chunks of a file is linked together as the children of the root. This tree is also stored as JSON data against the root's key.
```
joshua-kim, line 17: Diagram might be helpful here imo

docs/subnets/create-a-vm-blobvm.md
```diff
@@ -0,0 +1,1672 @@
...
+
+## High-Level Overview
+
+Virtual Machines are created as a module, whose binary is registered by a node running `avalanchego`, against the **vmID** (binary file name must be vmID). VMID is a user-defined string that is zero-extended to a 32-byte array and encoded in CB58.
+
+We can build multiple chains with the registered VM. A VM is registered, only when its built binary is present at the required location, as explained [here](#how-to-use-a-vm).
+
+### APIs for a VM
+
+The VM exposes 3 things for interaction:
```
joshua-kim, line 31: Wording here kind of suggests to me that clients are required as part of the VM interface, which isn't the case

docs/subnets/create-a-vm-blobvm.md
```diff
@@ -0,0 +1,1672 @@
...
+- **Static and Non-Static Handlers (API)** for making requests to VM methods
+- **Service** for implementing handlers or APIs
+- **Client** for making calls to the service's APIs
+
+### What are Handlers
+
+Handlers serve the response for the incoming HTTP requests. Handlers can also be wrapped with **gRPC** for efficiently making calls from other services such as `avalanchego`. Handler helps in creating APIs. VM implements 2 kinds of handlers:
+
+- **Non-Static Handlers** - They help in interacting with blockchains instantiated by the VM. The API's endpoint will be different for different chains. `/ext/bc/[chainID]`
+- **Static Handlers** - They help in directly accessing VM. These are optional for reasons such as parsing genesis bytes required to instantiate new blockchains. `/ext/vm/[vmID]`
```
joshua-kim, line 42: its a bit unclear what the difference between these two are from reading this

docs/subnets/create-a-vm-blobvm.md
joshua-kim: This feels like an unnecessary part of the doc imo, I would be fine removing it. We can just tell users that they need to implement the factory interface, but that's already called out in https://docs.avax.network/subnets/create-a-virtual-machine-vm#factory. I feel like letting users know what's going on under-the-hood is unnecessary for people trying to build VMs and will just be more stuff we need to update as the codebase evolves.

docs/subnets/create-a-vm-blobvm.md
```diff
@@ -0,0 +1,1672 @@
...
+### Instantiating a VM
+
+Each VM has a **factory** that is capable of creating new VM instances. A VM can be initialized as a chain along with handlers for accessing it, only when we have the VM's instance. The factory's `New` method shown below, returns the VM's instance to its caller in `avalanchego`. It's generally in the [`factory.go`](https://github.com/ava-labs/blobvm/blob/master/factory.go) file of the VM.
+
+```go
+// Returning a new VM instance from VM's factory
+func (f *Factory) New(*snow.Context) (interface{}, error) { return &vm.VM{}, nil }
+```
+
+A VM's functionality is exposed to `avalanchego` by registering its factory and [instantiating](https://github.com/ava-labs/avalanchego/blob/master/chains/manager.go#L399) a new VM using this factory whenever a chain is required to be built.
```
joshua-kim, line 66: IMO we can also remove this

docs/subnets/create-a-vm-blobvm.md
```diff
@@ -0,0 +1,1672 @@
...
+```go
+// Registering a factory inside avalanchego
+func (m *manager) RegisterFactory(vmID ids.ID, factory Factory) error {
+    ...
+    m.factories[vmID] = factory
+    ...
+}
+```
+
+### Instantiating a VM
```
joshua-kim, line 57: Feels like we're duplicating with the `timestampvm` tutorial in a lot of places in the doc [here](https://docs.avax.network/subnets/create-a-virtual-machine-vm#factory). Maybe we should break this + the timestampvm apart to have three docs? 
1. What needs to be implemented for a VM, explains each part of the vm interface and what it does
2. Timestamp tutorial, just references (1) and explains additional stuff we built on top to support timestampvm-specific features
3. Blob tutorial, similarly references (1) and explains additional stuff we built on top to support blobvm-specific features.

cc: @cgcardona @aaronbuchwald

docs/subnets/create-a-vm-blobvm.md
```diff
@@ -0,0 +1,1672 @@
...
+
+The main task of a VM is to issue blocks to the consensus engine whenever requested. It is the VM's responsibility to handle user transactions, and include them into new blocks, mempool handling, and database handling.
+
+The consensus engine will just request the VM for new blocks whenever there is a signal (from VM) and according to other [congestion control mechanisms](https://github.com/ava-labs/avalanchego/blob/master/vms/proposervm/README.md), verify the block, gossip the block within the network for consensus, and request the VM to accept or reject the block.
+
+Every VM should implement the following interfaces:
+
+### `block.ChainVM`
+
+To reach a consensus on linear blockchains (as opposed to DAG blockchains), Avalanche uses the Snowman consensus engine. To be compatible with Snowman, a VM must implement the `block.ChainVM` interface, which can be accessed from [AvalancheGo repository](https://github.com/ava-labs/avalanchego/blob/v1.7.4/snow/engine/snowman/block/vm.go).
```
joshua-kim, line 111: In the vein of the previous comment, I feel like we need to do some sort of doc refactor to eliminate duplicate info with https://docs.avax.network/subnets/create-a-virtual-machine-vm#blockchainvm.

joshua-kim (changes requested):

docs/subnets/create-a-vm-blobvm.md
rajranjan0608: Hi @joshua-kim. Thanks for this suggestion. We are moving ahead with this 3-tutorial VM series. I have made a recent commit to support this. Please have a look. Thank you.

docs/subnets/create-a-vm-blobvm.md
rajranjan0608: Initially, we were planning to replace the timestamp tutorial, that's why you were seeing duplications.

yulin-dong (dismissed):

docs/subnets/create-a-vm-blobvm.md
rajranjan0608: "etc" will be explained later in the tutorial

docs/subnets/introduction-to-vm.md
```diff
@@ -0,0 +1,361 @@
...
+    genesisData,
+    chainConfig.Upgrade,
+    chainConfig.Config,
+    msgChan,
+    fxs,
+    sender,
+);
+```
+
+You can refer to the [implementation](https://github.com/ava-labs/blobvm/blob/master/vm/vm.go#L92) of `vm.initialize` in the BlobVM repository.
```
yulin-dong, line 111: I would be helpful to provide the link to `vm.initialize` in the BlobVM repository?

docs/subnets/introduction-to-vm.md
```diff
@@ -0,0 +1,361 @@
...
+	// information about their accounts.
+	CreateHandlers() (map[string]*HTTPHandler, error)
+}
+```
+
+### `snowman.Block`
+
+You may have noticed the `snowman.Block` type referenced in the `block.ChainVM` interface. It describes the methods that a block must implement to be a block in a linear (Snowman) chain.
+
+Let’s look at this interface and its methods. You can see the full file from [here.](https://github.com/ava-labs/avalanchego/blob/v1.7.4/snow/consensus/snowman/block.go)
```
yulin-dong, line 282: Should the link to the master branch or v1.7.4 branch? If master branch, please search and replace in the doc.

docs/subnets/introduction-to-vm.md
```diff
@@ -0,0 +1,363 @@
...
+	// bytes.
+	ParseBlock([]byte) (snowman.Block, error)
+}
+```
+
+### `common.VM`
+
+`common.VM` is a type that every `VM`, whether a DAG or linear chain, must implement.
+
+You can see the full file from [here.](https://github.com/ava-labs/avalanchego/blob/v1.7.4/snow/engine/common/vm.go)
```
yulin-dong, line 193: should this link be to the master branch or v1.7.4? If master branch, please search and replace all v1.7.4.

docs/subnets/create-a-vm-blobvm.md
```diff
@@ -0,0 +1,1353 @@
...
+  - Get the execution context of the preferred block that includes
+    - Recent TxIDs (Txns in the lookback window e.g last 10s as defined in Genesis)
+    - Recent BlockIDs
+    - Recent Price and Cost
+    - Next Price, Next Cost, etc.
+  - Execute transaction locally with execution context, dummy database, and block
+  - Add valid transaction to mempool - `mempool.newTxs`
+    - Call `mempool.addPending()` to signal VM to build block for the newly added tx
+- Gossip new transactions from `mempool.newTxs` at regular intervals
+- Signals consensus engine to build blocks out of pending transactions in the mempool
```
yulin-dong, line 60: Please clarify who `Gossips` and `Signals`.

docs/subnets/create-a-vm-blobvm.md
```diff
@@ -0,0 +1,1353 @@
...
+- ProposerVM delays the request until it is the node's turn to propose a block
+- The consensus engine calls `vm.BuildBlock()` to get the block from VM
+- Engine calls `block.Verify()` method
+- Successfully verified blocks have gossiped within the network for consensus
```
yulin-dong, line 64:
```suggestion
- Successfully verified blocks have been gossiped within the network for consensus
```

yulin-dong (changes requested):

docs/subnets/create-a-vm-blobvm.md
```diff
@@ -0,0 +1,1353 @@
...
+- Blocks containing transactions are accepted or rejected according to the consensus results
+- Accepted blocks and related data are committed to the node's database
+
+## Coding the Virtual Machine
+
+We have divided the components into 3 packages. We will be looking at each of these files and learning about their functions.
+
+- **vm**
```
yulin-dong, line 72: Link to the directory on github project will be helpful.

docs/subnets/create-a-virtual-machine-vm.md
```diff
@@ -1452,7 +1180,7 @@ The path to the executable, as well as its name, can be provided to the build sc
 
 If no argument is given, the path defaults to a binary named with default VM ID: `$GOPATH/src/github.com/ava-labs/avalanchego/build/plugins/tGas3T58KzdjLHhBDMnH2TvrddhqTji5iZAMZ3RXs2NLpSnhH`
 
-This name `tGas3T58KzdjLHhBDMnH2TvrddhqTji5iZAMZ3RXs2NLpSnhH` is the CB58 encoded 32 byte identifier for the VM. For the timestampvm, this is the string "timestampvm" zero-extended in a 32 byte array and encoded in CB58.
+This name `tGas3T58KzdjLHhBDMnH2TvrddhqTji5iZAMZ3RXs2NLpSnhH` is the CB58 encoded 32 byte identifier for the VM. For the timestampvm, this is the string "timestampvm" zero-extended in a 32 byte array and encoded in CB58. See [here](./subnet-cli.md#subnet-cli-create-vmid) for more details.
```
joshua-kim, line 1183:
```suggestion
This name `tGas3T58KzdjLHhBDMnH2TvrddhqTji5iZAMZ3RXs2NLpSnhH` is the CB58 encoded 32 byte identifier for the VM. For the timestampvm, this is the string "timestampvm" zero-extended in a 32 byte array and encoded in CB58. See [here](./subnet-cli.md#subnet-cli-create-vmid) for more details on how to create a VM ID.
```

docs/subnets/introduction-to-vm.md
```diff
@@ -0,0 +1,363 @@
...
+### Handlers
+
+Handlers serve the response for the incoming HTTP requests. Handlers can also be wrapped with **gRPC** for efficiently making calls from other services such as `AvalancheGo`. Handlers help in creating APIs. VM implements 2 kinds of handlers:
+
+- **Non-Static Handlers** - They help in interacting with blockchains instantiated by the VM. The API's endpoint will be different for different chains. `/ext/bc/[chainID]`
+- **Static Handlers** - They help in directly accessing VM. These are optional for reasons such as parsing genesis bytes required to instantiate new blockchains. `/ext/vm/[vmID]`
+
+For any developers familiar with object-oriented programming, this is very similar to static and non-static methods on a class.
+
+### Registering a VM
```
joshua-kim, line 60: I think this was mentioned in a previous comment but I think this is info that's beyond the scope of this doc. We don't need to go into internals on how vms are registered/rpc calls/etc. People reading this doc are just trying to understand what VMs are/how to build one + install it.

docs/subnets/introduction-to-vm.md
```diff
@@ -0,0 +1,363 @@
...
+```go
+// Instantiating a new VM from its factory in avalanchego
+func (m *manager) buildChain(chainParams ChainParameters, sb Subnet) (*chain, error) {
+    ...
+    vmFactory, err := m.VMManager.GetFactory(vmID)
+    vm, err := vmFactory.New(ctx.Context)
+    ...
+}
+```
+
```
joshua-kim, line 95: This is internal to `avalanchego`, and won't be something users/vm builders will need to touch. We can just let them know that the vm factory interface is used to build the vm for avalanche go to consume. We can remove this section.

docs/subnets/introduction-to-vm.md
```diff
@@ -0,0 +1,363 @@
...
+	// returned.
+	Status() Status
+}
+```
+
+## rpcchainvm
+
+`rpcchainvm` is a special VM that wraps a `block.ChainVM` and allows the wrapped blockchain to run in its own process separate from AvalancheGo. `rpcchainvm` has two important parts: a server and a client. The [server](https://github.com/ava-labs/avalanchego/blob/master/vms/rpcchainvm/vm_server.go) runs the underlying `block.ChainVM` in its own process and allows the underlying VM's methods to be called via gRPC. The [client](https://github.com/ava-labs/avalanchego/blob/master/vms/rpcchainvm/vm_client.go) runs as part of `AvalancheGo` and makes gRPC calls to the corresponding server in order to update or query the state of the blockchain.
+
+To make things more concrete: suppose that `AvalancheGo` wants to retrieve a block from a chain run in this fashion. `AvalancheGo` calls the client's `GetBlock` method, which makes a gRPC call to the server, which is running in a separate process. The server calls the underlying VM's `GetBlock` method and serves the response to the client, which in turn gives the response to `AvalancheGo`.
```
joshua-kim, line 363: inline with previous comments I don't think this section is necessary

joshua-kim (changes requested):
Still reading the blobvm.md file, will post comments soon. Otherwise I think that this 3-part series looks much cleaner than the previous iteration, thanks for the docs!

docs/subnets/introduction-to-vm.md
rajranjan0608: Have referenced it in the `implementation` hyperlink.

docs/subnets/introduction-to-vm.md
rajranjan0608: This might be a little information that is cool to share, on how AvalancheGo can interact with the VM irrespective of its programming language.

docs/subnets/introduction-to-vm.md
```diff
@@ -0,0 +1,335 @@
...
+## Blocks and State
+
+Virtual Machines can be broken down into 2 components: blocks and state. The functionality provided by VMs is to:
+
+- Define the representation of a blockchain's state
+- Represent the operations in that state
+- Apply the operations in that state
+
+Each block in the blockchain contains a set of state transitions. Each block is applied in order from the blockchain's initial genesis block to its last accepted block to reach the latest state of the blockchain.
+## Blockchain
```
aaronbuchwald, line 22:
```suggestion

## Blockchain
```

docs/subnets/introduction-to-vm.md
```diff
@@ -0,0 +1,335 @@
...
+
+A blockchain has 2 components: **Consensus Engine** and **Virtual Machine**. VMs mainly deal with the implementation related to the block's structure: building and parsing. The Consensus Engine helps in reaching consensus on the block issued by VM. Here is a brief overview:
```
aaronbuchwald, line 24: nit: the VM does not issue a block the consensus engine does. The VM tells the consensus engine when it's ready to build a block and then the consensus engine calls `BuildBlock()`.

docs/subnets/introduction-to-vm.md
```diff
@@ -0,0 +1,335 @@
...
+
+1. A validator node wants to update the blockchain's state
```
aaronbuchwald, line 26: A non-validator can produce a block as well.

docs/subnets/introduction-to-vm.md
```diff
@@ -0,0 +1,335 @@
...
+2. It creates a block with state transition details in it
```
aaronbuchwald, line 27: Can we improve the phrase state transition details?

aaronbuchwald: We should be more specific with the subject of the sentence and never use something like `It`. We should be painfully specific here to make sure it's clear.

docs/subnets/introduction-to-vm.md
```diff
@@ -0,0 +1,335 @@
...
+3. It issues the block to the consensus engine
+4. Consensus engine gossips the block within the network to reach consensus on it
+5. Depending upon the consensus results, the engine can either accept or reject the block
+6. Every virtuous node on the network should have the same verdict for a particular block
+7. It depends on the VM implementation on what to do with the accepted or rejected block
+
+The consensus engine has been already provided by `AvalancheGo` on the Avalanche network. Developers can implement and customize their own VM.
```
aaronbuchwald, line 34: We should use the active voice as much as possible

```suggestion
AvalancheGo provides the consensus engine for every blockchain on the Avalanche Network. The consensus engine relies on the VM interface to handle building, parsing, and storing blocks as well as verifying and executing on behalf of the consensus engine.
```

docs/subnets/introduction-to-vm.md
```diff
@@ -0,0 +1,335 @@
...
+
+## How to Load a VM
+
+VMs are created as a module, whose binary is registered by a node running `AvalancheGo`, against the **vmID** (binary file name must be vmID). VMID is a user-defined string that is zero-extended to a 32-byte array and encoded in CB58.
```
aaronbuchwald, line 38: I guess this is true.. it's a 32 byte array that is cb58 encoded. It's not necessarily a user defined string although I suppose that's how we've used it. cc @StephenButtolph

docs/subnets/introduction-to-vm.md
```diff
@@ -0,0 +1,335 @@
...
+
+The Avalanche nodes that want to power their chains with a particular VM, must have the built binary put in the proper place. See [here](../nodes/maintain/avalanchego-config-flags.md#--build-dir-string) for more details. There could be multiple VM plugins in this directory.
```
aaronbuchwald, line 40: Can we clean this sentence up?

docs/subnets/introduction-to-vm.md
```diff
@@ -0,0 +1,335 @@
...
+
+We can build multiple blockchains with the registered VM. A VM is registered, only when its built binary is present at the required location.
```
aaronbuchwald, line 42: can we avoid the phrase `built binary` ? A lot of this phrasing is confusing.

docs/subnets/introduction-to-vm.md
```diff
@@ -0,0 +1,335 @@
...
+
+A blockchain can run as a separate process from AvalancheGo and can communicate with `AvalancheGo` over gRPC. This is enabled by `rpcchainvm`, a special VM that uses [`go-plugin`](https://pkg.go.dev/github.com/hashicorp/go-plugin) and wraps another VM implementation. The C-Chain, for example, runs the [Coreth](https://github.com/ava-labs/coreth) VM in this fashion.
```
aaronbuchwald, line 44: Note: we should state that rpcchainvm uses grpc. `go-plugin` relies on `grpc` and we're going to stop using `go-plugin` in the near future so might as well just say `grpc` instead.

docs/subnets/introduction-to-vm.md
```diff
@@ -0,0 +1,335 @@
...
+
+### APIs for a VM
+
+We interact with a blockchain and underlying VM through **Static** and **Non-Static Handlers**. Handlers are used for making API calls to VM methods. Specifically, these handlers are implemented as **Services**.
```
aaronbuchwald, line 48: Can we call them handlers and static handlers instead of static and non-static to coincide more directly with the function names on the VM?

I'm not a fan of calling them non-static as it makes static the default even though we barely use the static APIs.

aaronbuchwald: We should explain what a handler is in one sentence or call them HTTP handlers to make it more clear. Handler is a very vague term that can mean a lot of things.

docs/subnets/introduction-to-vm.md
```diff
@@ -0,0 +1,335 @@
...
+
+### Handlers
+
+Handlers serve the response for the incoming HTTP requests. Handlers can also be wrapped with **gRPC** for efficiently making calls from other services such as `AvalancheGo`. Handlers help in creating APIs. VM implements 2 kinds of handlers:
```
aaronbuchwald, line 52: Can we clean this paragraph up? I don't think the sentence `Handlers help in creating APIs` explains much.

docs/subnets/introduction-to-vm.md
```diff
@@ -0,0 +1,335 @@
...
+
+- **Non-Static Handlers** - They help in interacting with blockchains instantiated by the VM. The API's endpoint will be different for different chains. `/ext/bc/[chainID]`
+- **Static Handlers** - They help in directly accessing VM. These are optional for reasons such as parsing genesis bytes required to instantiate new blockchains. `/ext/vm/[vmID]`
+
+For any developers familiar with object-oriented programming, this is very similar to static and non-static methods on a class.
```
aaronbuchwald, line 57: Great callout!

docs/subnets/introduction-to-vm.md
```diff
@@ -0,0 +1,335 @@
...
+
+### VM Factory
+
+Each VM has a **factory** that is capable of creating new VM instances. A VM can be initialized as a blockchain along with handlers for accessing it, only when we have the VM's instance. The factory's `New` method shown below, returns the VM's instance to its caller in `AvalancheGo`. It's generally in the [`factory.go`](https://github.com/ava-labs/blobvm/blob/master/factory.go) file of the VM.
```
aaronbuchwald, line 61: Can we re-phrase here? This is hard to follow.

docs/subnets/introduction-to-vm.md
```diff
@@ -0,0 +1,335 @@
...
+
+You can refer to the [implementation](https://github.com/ava-labs/blobvm/blob/master/vm/vm.go#L92) of `vm.initialize` in the BlobVM repository.
+
+### Interaction
+
+We can use the API handlers to issue transactions, query chain state, and other functionalities provided by the VM.
+
+## Interfaces
+
+The main task of a VM is to issue blocks to the consensus engine whenever requested. It is the VM's responsibility to handle user transactions, and include them into new blocks, mempool handling, and database handling.
```
aaronbuchwald, line 93: I would not define this as the main task of the VM. The main task would be to verify blocks and handle accept/reject by performing the necessary state transitions.

docs/subnets/introduction-to-vm.md
aaronbuchwald: I agree this is valuable to include. Can we break this down into a step by step instead of a paragraph form to make it easier to follow?

docs/subnets/introduction-to-vm.md
aaronbuchwald, line 93: The main task of the VM is not to issue blocks to consensus, most VMs will not issue many blocks to consensus. The VM is responsible for parsing, storing, and building in no particular order and must also handle the verification and acceptance/rejecting of blocks and performing the corresponding state transitions.

docs/subnets/introduction-to-vm.md
```diff
@@ -0,0 +1,335 @@
...
+
+The consensus engine will just request the VM for new blocks whenever there is a signal (from VM) and according to other [congestion control mechanisms](https://github.com/ava-labs/avalanchego/blob/master/vms/proposervm/README.md), verify the block, gossip the block within the network for consensus, and request the VM to accept or reject the block.
```
aaronbuchwald, line 95: This paragraph does not fit under the interfaces header.

The content can use some cleaning up as well. Can we try:

ChainVMs are wrapped within the [ProposerVM]((https://github.com/ava-labs/avalanchego/blob/master/vms/proposervm/README.md)) to provide a soft leader. When a VM is ready to build a block, it will add a message to the `toEngine` channel that was passed in when the VM was initialized. The ProposerVM will intercept this message and wait for the time when it is that node's turn to propose a block on the chain to forward that message to the consensus engine. When the consensus engine receives that message, it will trigger a call to `BuildBlock()` on the VM, so that the VM can build a block to be issued into consensus.

Once the VM has built a block, the consensus engine will call verify on the block, and then issue it into consensus. The consensus engine actively drives consensus and whenever a block has been decided, the engine will call Accept/Reject on the block.

docs/subnets/create-a-virtual-machine-vm.md
```diff
@@ -1,294 +1,22 @@
...
-## Introduction
+- [Introduction to Virtual Machines](./introduction-to-vm.md)
+- How to Build a Simple VM (this article)
+- [How to Build a Complex VM](./create-a-vm-blobvm.md)
 
-One of the core features of Avalanche is the ability to create new, custom blockchains, which are defined by [Virtual Machines (VMs)](../overview/getting-started/avalanche-platform.md#virtual-machines)
+## Introduction
 
-In this tutorial, we’ll create a very simple VM. The blockchain defined by the VM is a [TimestampVM](https://github.com/ava-labs/timestampvm/tree/v1.2.1). Each block in the blockchain contains the timestamp when it was created along with a 32-byte piece of data (payload). Each block’s timestamp is after its parent’s timestamp.
+In this tutorial, we’ll create a very simple VM. The blockchain defined by the VM is the [TimestampVM](https://github.com/ava-labs/timestampvm/tree/v1.2.1). Each block in the TimestampVM's blockchain contains the timestamp when it was created along with a 32-byte piece of data. Each block’s timestamp is after its parent’s timestamp.
```
aaronbuchwald, line 11: This argument below about why the timestampvm is useful is a pretty tough sell. I think the purpose of this VM was as an example we don't really need to justify it anymore than that. Can we remove the paragraph below about how it will be useful for storing a book manuscript?

docs/subnets/introduction-to-vm.md
yulin-dong: Can we distinguish handlers as blockchain handlers and VM handlers instead? Non-static handlers are only accessible on a blockchain, and static handlers on VM.

rajranjan0608: Very nice. This sounds perfect to me ✌️.

yulin-dong: And I agree with Aaron that we should just call them APIs.

docs/subnets/introduction-to-vm.md
```diff
@@ -0,0 +1,337 @@
...
+
+Each block in the blockchain contains a set of state transitions. Each block is applied in order from the blockchain's initial genesis block to its last accepted block to reach the latest state of the blockchain.
+
+## Blockchain
+
+A blockchain has 2 components: **Consensus Engine** and **Virtual Machine**. VMs mainly deal with the implementation related to the block's structure: building and parsing. The Consensus Engine helps in reaching consensus on the block built by the VM. Here is a brief overview:
+
+1. A node wants to update the blockchain's state
+2. The node's VM will notify the consensus engine that it wants to update the state
+3. Consensus engine will request the block from the VM with updates in it
```
yulin-dong, line 29:
```suggestion
3. Consensus engine will request the block from the VM with updates
```

docs/subnets/introduction-to-vm.md
```diff
@@ -0,0 +1,337 @@
...
+4. Consensus engine will verify the returned block. It will rely on VM's implementation of `verify()` method
```
yulin-dong, line 30:
```suggestion
4. Consensus engine will verify the returned block and rely on VM's implementation of `verify()` method
```

docs/subnets/introduction-to-vm.md
```diff
@@ -0,0 +1,337 @@
...
+5. Consensus engine will gossip the successfully verified blocks within the network to reach consensus on it
```
yulin-dong, line 31:
```suggestion
5. Consensus engine will gossip the successfully verified blocks within the network to reach consensus on this block
```

docs/subnets/introduction-to-vm.md
```diff
@@ -0,0 +1,337 @@
...
+6. Depending upon the consensus results, the engine can either accept or reject the block
+7. Every virtuous node on the network should have the same verdict for a particular block
+8. It depends on the VM implementation on what to do with the accepted or rejected block
```
yulin-dong, line 34: Can you clarify what `it` is in this sentence?

docs/subnets/introduction-to-vm.md
```diff
@@ -0,0 +1,337 @@
...
+Each VM has a **factory** that is capable of creating new VM instances. A VM can be initialized as a blockchain along with handlers for accessing it, only when we have the VM's instance. The factory's `New` method shown below, returns the VM's instance to its caller in `AvalancheGo`. It's generally in the [`factory.go`](https://github.com/ava-labs/blobvm/blob/master/factory.go) file of the VM.
+
+```go
+// Returning a new VM instance from VM's factory
+func (f *Factory) New(*snow.Context) (interface{}, error) { return &vm.VM{}, nil }
+```
+
+### Initializing a VM
+
+Blockchains are functional, only when the instantiated VMs are initialized and the node is bootstrapped. Initializing a VM involves setting up the database, block builder, mempool, genesis state, handlers, etc. This will expose the VM's API handlers, start accepting transactions, gossiping them across the network, building blocks, etc. More details on it can be found in the [third part](/subnets/create-a-vm-blobvm#initialize) of this series.
```
yulin-dong, line 72: Should `This will expose the VM's API handlers` be changed to `This will expose the blockchain's API handlers` ?

docs/subnets/introduction-to-vm.md
```diff
@@ -0,0 +1,333 @@
...
+
+- Define the representation of a blockchain's state
+- Represent the operations in that state
+- Apply the operations in that state
+
+Each block in the blockchain contains a set of state transitions. Each block is applied in order from the blockchain's initial genesis block to its last accepted block to reach the latest state of the blockchain.
+
+## Blockchain
+
+A blockchain has 2 components: **Consensus Engine** and **Virtual Machine**. VMs mainly deal with the implementation related to the block's structure: building and parsing. The Consensus Engine helps in reaching consensus on the block built by the VM. Here is a brief overview:
```
yulin-dong, line 25: Do you have a circular definition? A VM can initiate multiple blockchains. Here we say `A blockchain has 2 components: **Consensus Engine** and **Virtual Machine**.`. We need a better explanation.

docs/subnets/introduction-to-vm.md
```diff
@@ -0,0 +1,333 @@
...
+
+In order to load a VM, its binary must be put it the proper place where `AvalancheGo` can find and load it. See [here](../nodes/maintain/avalanchego-config-flags.md#--build-dir-string) for more details. There could be multiple VM plugins in this directory.
+
+A VM can be initialized as blockchains. A blockchain can run as a separate process from AvalancheGo and can communicate with `AvalancheGo` over gRPC. This is enabled by `rpcchainvm`, a special VM that uses [`go-plugin`](https://pkg.go.dev/github.com/hashicorp/go-plugin) and wraps another VM implementation. The C-Chain, for example, runs the [Coreth](https://github.com/ava-labs/coreth) VM in this fashion.
+
+### API Handlers
+
+We interact with a blockchain and underlying VM through API handlers. Specifically, these handlers are implemented as **Services**.
+
+Handlers serve the response for the incoming HTTP requests; and handlers can also be wrapped with **gRPC** for efficiently making calls from other services such as `AvalancheGo`. VM implements 2 kinds of handlers:
```
yulin-dong, line 50: `with **gRPC** for efficiently making calls from other services` or `with **gRPC** for efficiently responding to calls from other services` ?

docs/subnets/create-a-vm-blobvm.md
```diff
@@ -0,0 +1,1353 @@
...
+- **Builder** - Returns the built preferred block by including transactions from the mempool.
+- **Block Builder** - Initiates the gossip of new transactions, notifies the engine that new blocks are ready to be built, and also handles next block generation notification.
+- **Storage** - Stores and retrieves data of the chain's state.
+- **Virtual Machine** - Entry point for all the components to orchestrate them according to the consensus engine and API requests.
+- **Service** - API handlers for interacting with VM and initialized chain.
+- **Factory** - For creating new instances of the Virtual Machine.
+
+## Transaction Lifecycle in BlobVM
+
+A Virtual Machine exposes APIs or handlers for users to make direct RPC to service or use the client to interact with the service. Every change on a chain happens through blocks and more specifically, transactions. A VM handles transactions internally since the consensus engine only cares about the block. Let's see, how a transaction goes through the network to update the chain's state:
```
yulin-dong, line 41: what does the client in `use the client` refer to?

docs/subnets/create-a-vm-blobvm.md
```diff
@@ -0,0 +1,1353 @@
...
+	if b.Price < g.MinPrice {
+		return ErrInvalidPrice
+	}
+	return nil
+}
+```
+
+- [`Execute()`](https://github.com/ava-labs/blobvm/blob/master/chain/unsigned_tx.go#L34) executes the specific check for a transaction and may perform state change on the database instance provided as an argument. Each type of transaction should implement its own execute method. For eg. `TransferTx` execute balance modification i.e. add transfer amount to the receiver and deduct the same amount from the sender.
+
+A transaction is executed 2 times. Before [including](https://github.com/ava-labs/blobvm/blob/master/vm/vm.go#L428) it in mempool and during [verification](https://github.com/ava-labs/blobvm/blob/master/chain/block.go#L213) of the block containing this transaction. The database for the former is aborted as this is just a local execution for validating transactions before gossiping. Whereas the database for the latter is committed after the block is accepted by the network.
```
yulin-dong, line 159: This paragraph is not easy to understand. I would suggest to rewrite it.

rajranjan0608: Added a few clarifications.

docs/subnets/create-a-vm-blobvm.md
rajranjan0608: I should remove the `client` reference, as this is not a requirement for BlobVM.

docs/subnets/introduction-to-vm.md
rajranjan0608: How about this - 

with **gRPC** for efficiently communicating with other services...

docs/subnets/introduction-to-vm.md
rajranjan0608: Can we use this - 

A blockchain relies on 2 components: **Consensus Engine** and **Virtual Machine**. VM is the blueprint for a blockchain that mainly deals with...

yulin-dong: Sounds good to me.

docs/subnets/introduction-to-vm.md
yulin-dong: OK. For my understanding, are Handlers making call to or being called from other services?

docs/subnets/create-a-vm-blobvm.md
```diff
@@ -0,0 +1,1359 @@
...
+	if b.Price < g.MinPrice {
+		return ErrInvalidPrice
+	}
+	return nil
+}
+```
+
+- [`Execute()`](https://github.com/ava-labs/blobvm/blob/master/chain/unsigned_tx.go#L34) executes the specific check for a transaction and may perform state change on the database instance provided as an argument. Each type of transaction should implement its own execute method. For example, `TransferTx` execute balance modification, i.e. add transfer amount to the receiver and deduct the same amount from the sender.
+
+A transaction is executed 2 times. Before [including](https://github.com/ava-labs/blobvm/blob/master/vm/vm.go#L428) it in mempool and during [verification](https://github.com/ava-labs/blobvm/blob/master/chain/block.go#L213) of the block containing this transaction. As mentioned earlier, verification of a block happens before gossiping it within the network for consensus.
```
yulin-dong, line 159:
```suggestion
A transaction is executed 2 times: before [including](https://github.com/ava-labs/blobvm/blob/master/vm/vm.go#L428) it in mempool and during [verification](https://github.com/ava-labs/blobvm/blob/master/chain/block.go#L213) of the block containing this transaction. As mentioned earlier, verification of a block happens before gossiping it within the network for consensus.
```

docs/subnets/introduction-to-vm.md
rajranjan0608: AFAIK, handlers are being called from other services (here avalanchego)

docs/subnets/introduction-to-vm.md
```diff
@@ -0,0 +1,341 @@
...
+Each VM has a **factory** that is to create new VM instances from which a blockchain can be initialized. The factory's `New` method as shown below, returns the VM's instance to its caller in `AvalancheGo`. It's generally in the [`factory.go`](https://github.com/ava-labs/blobvm/blob/master/factory.go) file of the VM.
+
+```go
+// Returning a new VM instance from VM's factory
+func (f *Factory) New(*snow.Context) (interface{}, error) { return &vm.VM{}, nil }
+```
+
+### Initializing a VM to Create a Blockchain
+
+Blockchains are functional, only when the instantiated VMs are initialized and the node has been bootstrapped. Initializing a VM involves setting up the database, block builder, mempool, genesis state, handlers, etc. This will expose the VM's API handlers, start accepting transactions, gossiping them across the network, building blocks, etc. More details on it can be found in the [third part](./create-a-vm-blobvm.md#initialize) of this series.
```
joshua-kim, line 68:
```suggestion
Before a VM can run, AvalancheGo will initialize it by invoking its `Initialize` method. Here, the VM will bootstrap itself and sets up anything it requires before it starts running.

This might involve setting up its database, mempool, genesis state, or anything else the VM requires to run. For more information, see the [third part](./create-a-vm-blobvm.md#initialize) of this series.
```

docs/subnets/introduction-to-vm.md
```diff
@@ -0,0 +1,341 @@
...
+    fxs,
+    sender,
+);
+```
+
+You can refer to the [implementation](https://github.com/ava-labs/blobvm/blob/master/vm/vm.go#L92) of `vm.initialize` in the BlobVM repository.
+
+### Interaction
+
+We can use the API handlers to issue transactions, query chain state, and other functionalities provided by the blockchain.
```
joshua-kim, line 87:
```suggestion
```

docs/subnets/introduction-to-vm.md
```diff
@@ -0,0 +1,341 @@
...
+
+### ProposerVM and Consensus Engine
+
+ChainVMs are wrapped within the [ProposerVM](https://github.com/ava-labs/avalanchego/blob/master/vms/proposervm/README.md) to provide a soft leader. When a VM is ready to build a block, it will add a message to the `toEngine` channel that was passed in when the VM was initialized. The ProposerVM will intercept this message and wait for the time when it is that node's turn to propose a block on the chain to forward that message to the consensus engine. When the consensus engine receives that message, it will trigger a call to `BuildBlock()` on the VM, so that the VM can build a block to be issued into consensus.
+
+Once the VM has built a block, the consensus engine will call verify on the block, and then issue it into consensus. The consensus engine actively drives consensus and whenever a block has been decided, the engine will call Accept/Reject on the block.
```
joshua-kim, line 93:
```suggestion
```

docs/subnets/introduction-to-vm.md
```diff
@@ -0,0 +1,341 @@
...
+
+## Interfaces
+
+As stated before, the VM is responsible for parsing, storing, and building blocks (in no particular order) and must also handle the verification, acceptance/rejection of blocks and performing the corresponding state transitions.
+
```
joshua-kim, line 98:
```suggestion
```

docs/subnets/introduction-to-vm.md
```diff
@@ -0,0 +1,341 @@
...
+Every VM should implement the following interfaces:
+
+### `block.ChainVM`
+
+To reach a consensus on linear blockchains (as opposed to DAG blockchains), Avalanche uses the Snowman consensus engine. To be compatible with Snowman, a VM must implement the `block.ChainVM` interface, which can be accessed from [AvalancheGo repository](https://github.com/ava-labs/avalanchego/blob/master/snow/engine/snowman/block/vm.go).
```
joshua-kim, line 103:
```suggestion
To reach a consensus on linear blockchains, Avalanche uses the Snowman consensus engine. To be compatible with Snowman, a VM must implement the `block.ChainVM` interface. For more information, see [here](https://github.com/ava-labs/avalanchego/blob/master/snow/engine/snowman/block/vm.go).
```

docs/subnets/introduction-to-vm.md
```diff
@@ -0,0 +1,341 @@
...
+	//
+	// The block should be represented by the full byte array, without extra
+	// bytes.
+	ParseBlock([]byte) (snowman.Block, error)
+}
+```
+
+### `common.VM`
+
+`common.VM` is a type that every `VM`, whether a DAG or linear chain, must implement.
```
joshua-kim, line 165:
```suggestion
`common.VM` is a type that every `VM` must implement.
```

docs/subnets/introduction-to-vm.md
```diff
@@ -0,0 +1,341 @@
...
+
+You can see the full file from [here.](https://github.com/ava-labs/avalanchego/blob/master/snow/engine/common/vm.go)
```
joshua-kim, line 167:
```suggestion
For more information, you can see the full file [here](https://github.com/ava-labs/avalanchego/blob/master/snow/engine/common/vm.go).
```

docs/subnets/introduction-to-vm.md
```diff
@@ -0,0 +1,341 @@
...
+	// For example, if this VM implements an account-based payments system,
+	// it have an extension called `accounts`, where clients could get
+	// information about their accounts.
+	CreateHandlers() (map[string]*HTTPHandler, error)
+}
+```
+
+### `snowman.Block`
+
+You may have noticed the `snowman.Block` type referenced in the `block.ChainVM` interface. It describes the methods that a block must implement to be a block in a linear (Snowman) chain.
```
joshua-kim, line 256:
```suggestion
The `snowman.Block` interface It define the functionality a block must implement to be a block in a linear Snowman chain.
```

docs/subnets/introduction-to-vm.md
```diff
@@ -0,0 +1,341 @@
...
+
+Let’s look at this interface and its methods. You can see the full file from [here.](https://github.com/ava-labs/avalanchego/blob/master/snow/consensus/snowman/block.go)
```
joshua-kim, line 258:
```suggestion
For more information, you can see the full file [here](https://github.com/ava-labs/avalanchego/blob/master/snow/consensus/snowman/block.go).
```

docs/subnets/introduction-to-vm.md
```diff
@@ -0,0 +1,341 @@
...
+    Bytes() []byte
+
+    // Height returns the height of this block in the chain.
+    Height() uint64
+}
+```
+
+### `choices.Decidable`
+
+This interface is the superset of every decidable object, such as transactions, blocks, and vertices. You can see the full file from [here.](https://github.com/ava-labs/avalanchego/blob/master/snow/choices/decidable.go)
```
joshua-kim, line 297:
```suggestion
This interface is a superset of every decidable object, such as transactions, blocks, and vertices. For more information, you can see the full file [here](https://github.com/ava-labs/avalanchego/blob/master/snow/choices/decidable.go).
```

docs/subnets/introduction-to-vm.md
```diff
@@ -0,0 +1,341 @@
...
+	// If Accept has been called on an element with this ID, Accepted should be
+	// returned. Similarly, if Reject has been called on an element with this
+	// ID, Rejected should be returned. If the contents of this element are
+	// unknown, then Unknown should be returned. Otherwise, Processing should be
+	// returned.
+	Status() Status
+}
+```
+
+## rpcchainvm
```
joshua-kim, line 333: This seems like duplicate from the earlier rpcchainvm section

docs/subnets/introduction-to-vm.md
```diff
@@ -0,0 +1,335 @@
+# Introduction to Virtual Machines
+
+## Overview
+
+This is part of a series of tutorials for building a Virtual Machine (VM):
+
+- Introduction to Virtual Machines (this article)
```
joshua-kim, line 7:
```suggestion
This is part of a series of tutorials for building a Virtual Machine (VM):

- Introduction to VMs (this article)
```

docs/subnets/introduction-to-vm.md
```diff
@@ -0,0 +1,335 @@
+# Introduction to Virtual Machines
```
joshua-kim, line 1:
```suggestion
# Introduction to VMs
```

docs/subnets/create-a-vm-blobvm.md
```diff
@@ -0,0 +1,1359 @@
...
+
+This is part of a series of tutorials for building a Virtual Machine (VM):
+
+- [Introduction to Virtual Machines](./introduction-to-vm.md)
+- [How to Build a Simple VM](./create-a-vm-timestampvm.md)
+- How to Build a Complex VM (this article)
+
+## Introduction
+
+In this tutorial, we will learn all about how we can build a virtual machine by referencing [BlobVM](https://github.com/ava-labs/blobvm). It's a Virtual Machine that can be used to instantiate key-value blockchains for storing files like images, videos, etc. in an efficient way.
```
aaronbuchwald, line 11: I don't think it's true that BlobVM will scale well for images/videos, so we should probably remove those references.

joshua-kim (approved):
lgtm cc @aaronbuchwald

docs/subnets/create-a-vm-blobvm.md
joshua-kim: I removed this reference and just said generic k/v storage.

Totally an aside but I think it kind of does scale fine but it would depend on the use-case/if you consider gas prices an issue in scaling :p

yulin-dong (approved):

PR #893, opened 2022-07-08
[AV-1850] update config flags to match geth v1.10.20 changes

darioush:
merge with next public release of AvalancheGo


aaronbuchwald (approved):
LGTM - this should be merged with the next coreth/avalanchego release.

docs/nodes/maintain/chain-config-flags.md
```diff
@@ -122,20 +122,47 @@ Specifies the directory for the Admin API to use to store CPU/Mem/Lock Profiles.
...
+
+
 :::note
 
 If you populate this field, it will override the defaults so you must include every service you wish to enable.
 
 :::
 
-#### `eth`:
+#### `public-eth` (deprecated) or `eth`:
```
aaronbuchwald, line 158: @yulin-dong how do we normally handle deprecation notices like this?

It seems strange to keep the deprecated name in the header here. Should we move the old name and deprecation notice into the body and say something like:

```
The API name`public-eth` is deprecated as of v1.7.15, and the APIs previously under this name have been migrated to `eth`.
```

yulin-dong: Yeah, I like your suggestion better, @aaronbuchwald

yulin-dong (approved):

yulin-dong:
I will merge it to branch 1.7.5 which will be merged to master when we are ready.

PR #906, opened 2022-07-16
Added project branch

FibrinLab:
Awaiting review so I can implement corrections.


jpop32:
Hi there!

Thank you for your submission, and congratulations on making it to Round 2! In this round, the idea is to polish the tutorial and make certain changes to make it more accessible, more in line with our needs and more uniform in looks with the rest of the material we have on our docs pages.

Once other reviewers from our team have had a chance to look over the material in more detail they will post their suggestions as well, but for now here are some general guidelines to check for:
* large (more than a dozen or lines or so) source files should be linked to, not embedded. It's ok to highlight outtakes of the most important parts of the file, but just presenting the whole file so that it can be copied is best avoided.
* alerts, notes and warnings should be done using Markdown tags as shown [here](https://github.com/ava-labs/avalanche-docs/blob/master/docs/nodes/maintain/node-backup-and-restore.md#nodeid).
* blocks of text/code/data should be annotated with proper style tags like text, bash, json and so on. Check [here](https://raw.githubusercontent.com/ava-labs/avalanche-docs/master/docs/nodes/build/set-up-node-with-installer.md)
for examples.

But, first, you will probably need to redo the PR because we don't see the submission files inside, can you check what's up with that? Please refer to other submission PRs (marked with `subnet doc contest` tag).

FibrinLab:
Done

PR #1028, opened 2022-09-26
Expanded and reformatted the Subnet vs C-Chain article

jpop32:


gyuho (approved):

yulin-dong (approved):

docs/subnets/when-to-use-subnet-vs-c-chain.md
```diff
@@ -1,33 +1,58 @@
...
 
-## We Want Our Own Gas Token
+## Reasons To Use a Subnet
+
+There are many advantages to running your own Subnet. If you find one or more of these a good match for your project then a Subnet might be a good solution for you. But make sure to check the reasons to use th C-Chain instead, as some trade-offs involved might that a preferred solution.
```
darioush, line 7:
```suggestion
There are many advantages to running your own Subnet. If you find one or more of these a good match for your project then a Subnet might be a good solution for you. But make sure to check the reasons to use the C-Chain instead, as some trade-offs involved might make that a preferred solution.
```

PR #1063, opened 2022-11-06
Update README.md

eino-makitalo:
This is unfortunately not true as we know. All subnets are in the same directory in the operating system. And as they always validate also with the primary (P) network, this is a MUST condition.  If something goes wrong in the subnet, you have to clear the whole database.  There is no way to rest only the subnet and restart it from the beginning.

### DIFF
--- a/docs/subnets/README.md
+++ b/docs/subnets/README.md
@@ -10,7 +10,7 @@ By definition, all Subnet validators must also validate the Avalanche Primary Ne
 
 (Image adopted from [this article](https://www.coinbase.com/cloud/discover/dev-foundations/intro-to-avalanche-subnets))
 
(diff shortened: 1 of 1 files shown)


docs/subnets/README.md
```diff
@@ -69,3 +69,6 @@ Avalanche VMs (AVMs) make it easy to define a blockchain-based decentralized app
 ## Developing Your Own Subnet
 
 Please check out documents listed on the left panel to develop your own Subnets with customized virtual machines and blockchains.
+
+## Notes
+(1) Subnets do not share storage in logical level (keys/values) with other subnets, but they share storage on disk level (LevelDB) and store their data into same database/folder in operating system.
```
yulin-dong, line 74:
```suggestion
(1) Subnets do not share storage in logical level (keys/values) with other Subnets, but they share storage on disk level (LevelDB) and store their data into same database/folder in operating system.
```

yulin-dong (changes requested):

yulin-dong (approved):

PR #1080, opened 2022-11-23
Small Corrections to Hello World Precompile Tutorial

anusha-ctrl:


docs/subnets/hello-world-precompile-tutorial.md
```diff
@@ -867,7 +867,7 @@ Going to `./contract-examples`, we can finally run our tests by running the comm
 npx hardhat test --network local
 ```
 
-Great they passed! All the functions implemented in the precompile work as expected!
+Great the `Hello World` tests passed! All the functions implemented in the precompile work as expected! Other precompile tests will fail as their precompiles are not enabled. We only enabled the `Hello World` precompile in the genesis!
```
aaronbuchwald, line 870: Can we modify the command that is run so that users will only run the tests that are expected to pass?

It's not a great UX for people going through the tutorial to see a bunch of failing unit tests even if there is a clear explanation, so a workaround to run a specific set of tests would be very useful here.

Maybe using the grep option: https://ethereum.stackexchange.com/questions/113296/run-a-single-hardhat-test

aaronbuchwald (approved):
LGTM - one comment for future changes to improve test output

docs/subnets/hello-world-precompile-tutorial.md
anusha-ctrl: Sounds good. Will add it in the next round.

PR #1083, opened 2022-11-25
Correct ANR Installation Command

anusha-ctrl:
Quick correction to installation instructions for ANR. 

 I've gotten some feedback that some of the installation instructions are wrong or unclear.
For example
`go install -v ./cmd/avalanche-network-runner`
should be
`go install -v ./cmd/`

Removed those instructions and added the standard curl command.

### DIFF
--- a/docs/subnets/network-runner.md
+++ b/docs/subnets/network-runner.md
@@ -28,7 +28,7 @@ To build and install the binary locally (requires `golang` to be installed. Chec
 
 ```bash
 cd ${HOME}/go/src/github.com/ava-labs/avalanche-network-runner
-go install -v ./cmd/avalanche-network-runner
+go install -v ./cmd
 ```
 
 `avalanche-network-runner` will be installed into `$GOPATH/bin`, please make sure that `$GOPATH/bin` is in your `$PATH`, otherwise, you may not be able to run commands below.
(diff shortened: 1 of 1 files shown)


docs/subnets/network-runner.md
```diff
@@ -28,7 +28,7 @@ To build and install the binary locally (requires `golang` to be installed. Chec
 
 ```bash
 cd ${HOME}/go/src/github.com/ava-labs/avalanche-network-runner
-go install -v ./cmd/avalanche-network-runner
+go install -v ./cmd
```
felipemadero, line 31: could you pls instead refer to using the install.sh script, instead of `go install`, as using the script results in a bin that can be asked for versioning information (--version flag)
```
curl -sSfL https://raw.githubusercontent.com/ava-labs/avalanche-network-runner/main/scripts/install.sh | sh -s
```
for install under `./bin/` (need to change text afterwards)
or maybe better
```
curl -sSfL https://raw.githubusercontent.com/ava-labs/avalanche-network-runner/main/scripts/install.sh | sh -s -- -b $GOPATH/bin
```
so text afterwards does not need to be changed

felipemadero: I noticed text before the command should be changed, probably:
- remove everything from Installation title until the command, instead write something like `To install the latest binary locally`, and probably a note that `GOPATH` should be defined
- put second curl command (one that writes into `$GOPATH/bin`), replacing both the cd and the go install

anusha-ctrl: done thanks @felipemadero

cgcardona (approved):

PR #1116, opened 2022-12-08
add subnet uptimes docs

ceyonur:
Adds subnet uptime related API changes

### DIFF
--- a/.markdownlint.json
+++ b/.markdownlint.json
@@ -1,6 +1,10 @@
 {
   "default": true,
-  "MD013": { "line_length": 100 },
+  "MD013": {
+    "line_length": 100,
+    "code_blocks": false,
+    "tables": false
+  },
   "MD033": false,
   "no-hard-tabs": false,
   "whitespace": false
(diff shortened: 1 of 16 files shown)


.markdownlint.json
```diff
@@ -1,7 +1,11 @@
 {
   "default": true,
-  "MD013": { "line_length": 100 },
+  "MD013": {
+    "line_length": 100,
+    "code_blocks": false,
+    "tables": false
+  },
```
yulin-dong, line 7: @dasconnor I am not familiar with this. Please comment.

yulin-dong (changes requested):

.markdownlint.json
dasconnor: ahh, this is a good change. These are the settings for the line length check. We used to need to manually ignore code blocks, so this will make it so those are automatically exempt. Also excludes tables. I'm good with this.

yulin-dong (approved):

PR #1150, opened 2023-01-04
Redo home page

dasconnor:
I redid our home page to be more tile focused. Definitely open to feedback on links and copy. I'd like to replace the API link with a page on warp messaging when we have a better doc up.

### DIFF
--- a/src/pages/index.js
+++ b/src/pages/index.js
@@ -11,83 +11,85 @@ function Home() {
   return (
     <Layout title="Homepage" description="Avalanche Documentation">
       <main>
+        <br/>
+        <h1 align="center" style={{ fontWeight: '750' }}>Welcome to Avalanche Docs</h1>
         <section className={styles.features}>
           <div className="container">
             <div className="row cards__container">
               <Card
-                to="developers"
+                to="subnets/build-first-subnet"
                 header={{
-                  label: "⛏ Start Developing",
+                  label: "🚀 Launch Your First Subnet",
                   translateId: "start-building",
                 }}
                 body={{
                   label:
-                    "Get started building your decentralized app or marketplace with Solana.",
+                    "Start your Subnet development journey by creating a subnet in under five minutes",
                   translateId: "get-started-building",
                 }}
               />
 
               <Card
-                to="running-validator"
+                to="intro"
                 header={{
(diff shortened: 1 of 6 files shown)


src/pages/index.js
```diff
@@ -0,0 +1,92 @@
...
+      <main>
+        <br/>
+        <h1 align="center" style={{ fontWeight: '750' }}>Welcome to Avalanche Docs</h1>
+        <section className={styles.features}>
+          <div className="container">
+            <div className="row cards__container">
+              <Card
+                to="subnets/build-first-subnet"
+                header={{
+                  label: "🚀 Launch Your First Subnet",
```
yulin-dong, line 22: Not this PR, however, we would need to consider how to i18n'ize these strings.

yulin-dong (changes requested):

src/pages/index.js
```diff
@@ -0,0 +1,92 @@
...
+                  label: "🛠️ Launch Your Dapp on Avalanche",
+                }}
+                body={{
+                  label:
+                    "Learn everything you need to deploy an EVM-compatible smart contract",
+                }}
+              />
+
+              <Card
+                to="https://core.app"
```
yulin-dong, line 75: Not sure about this link to core.app, as it does not do an explanation what Core is. https://support.avax.network/en/collections/3391518-core is not a good choice, either.

dasconnor: I was thinking we wanted a link to each of our most important topics, so I wanted to have one for Core. Not sure what the best page that represents Core is, but that one seemed like the best I could find.

src/pages/index.js
```diff
@@ -0,0 +1,92 @@
...
+                body={{
+                  label:
+                    "Discover how Subnets and Avalanche Consensus are revolutionizing Web3",
+                }}
+              />
+
+              <Card
+                to="nodes"
+                header={{
+                  label: "😎 Become a validator",
```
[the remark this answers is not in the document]
dasconnor: fixed this

yulin-dong (approved):

src/pages/index.js
dasconnor: going to leave as is for now, happy to update later if we have more consensus

PR #1179, opened 2023-01-26
Remove mostly empty key features section from intro

aaronbuchwald:
This PR removes the mostly empty key features section from the docs intro page.

This currently looks quite bad, so this is a first step to improving the intro page.


cgcardona (approved):
LGTM

PR #1190, opened 2023-02-01
Fix the platform.getBalance `address` param

cgcardona:
Fix the example `curl`. Previously it was passing in a single address as `address`. Now it's correctly passing in an array of strings as `addresses`


yulin-dong:
LGTM

PR #1196, opened 2023-02-05
Update PrecompileGen tutorial to use create precompile script

aaronbuchwald:
This PR updates the PrecompileGen tutorial to use the new script `./scripts/create_precompile.sh` instead of directly running it with the `go run ...` command.

This change was made to address https://github.com/ava-labs/subnet-evm/issues/483.

Since we use the blst library in AvalancheGo, we need to set a CGO flag when compiling/running subnet-evm on some OSs. I have not been able to reproduce this bug, so we should hold off on merging this PR until the reporter confirms that this resolves the issue and the corresponding subnet-evm PR gets merged.

### DIFF
--- a/docs/subnets/hello-world-precompile-tutorial.md
+++ b/docs/subnets/hello-world-precompile-tutorial.md
@@ -274,7 +274,7 @@ Let's go back to the root of the Subnet-EVM repository and run the PrecompileGen
 ```bash
 cd $GOPATH/src/github.com/ava-labs/subnet-evm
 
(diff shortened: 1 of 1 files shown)


ceyonur (changes requested):
I commented in the feature-PR

https://github.com/ava-labs/subnet-evm/pull/489#issuecomment-1419227290

darioush (approved):

anusha-ctrl (approved):

yulin-dong (approved):

ceyonur:
Hmm I think we should have waited until we have a release for subnet-evm? What do you say? @aaronbuchwald

PR #1265, opened 2023-03-17
Add subnet checklist

chand1012:
Hey all! @gatesyp and I thought a subnet checklist would be helpful on the docs.


dasconnor:
Appreciate the work! Would you mind moving this under our community page? There's a few recommendations here that I don't think we can make in the "official" docs. Also, there are a few linting errors to take care of. Thanks!

PR #1311, opened 2023-04-19
Update docusaurus.config.js

dubrado:


yulin-dong (changes requested):
Please update the package.json to replace https://docusaurus.io/docs/api/plugins/@docusaurus/plugin-google-analytics with https://docusaurus.io/docs/api/plugins/@docusaurus/plugin-google-gtag

PR #1336, opened 2023-05-03
Adding an info box regarding SSH git clone 

laviniat1996:
We use SSH git cloning, and not all users are aware of this stuff. Recently I've seen a user having trouble cloning the repo because of this.

I suggest adding this warning message that let's users know about SSH and the HTTPS alternative.


meaghanfitzgerald (approved):
lgtm

PR #1365, opened 2023-05-24
Update subnet interaction needs

dubrado:


dubrado:
HOLD for review from dev for accuracy.

PR #1372, opened 2023-05-30
Update chain-config-flags.md

dubrado:
clarification for pruning-enabled

### DIFF
--- a/docs/nodes/maintain/chain-config-flags.md
+++ b/docs/nodes/maintain/chain-config-flags.md
@@ -631,8 +631,9 @@ Defaults to `4096`.
 
 #### `pruning-enabled` (boolean)
 
-If `true`, database pruning of obsolete historical data will be enabled. This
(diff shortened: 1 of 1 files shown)


docs/nodes/maintain/chain-config-flags.md
```diff
@@ -631,8 +631,9 @@ Defaults to `4096`.
 
 #### `pruning-enabled` (boolean)
 
-If `true`, database pruning of obsolete historical data will be enabled. This
-flag should be set to `false` for nodes that need access to all data at
+If `true`, database pruning of obsolete historical data will be enabled. This reduces the amount 
+of data written to disk, but does not delete any state that is written to the disk.
```
jpop32, line 635: Maybe '...delete any existing state that was written to the disk previously.' ?

dubrado: Good point, added.

amandamarquis1 (approved):

PR #1386, opened 2023-06-06
Change cloning method to HTTPS

laviniat1996:
Changed the cloning method to HTTPS for ease of use. Added an info box about SSH.

### DIFF
--- a/docs/nodes/build/run-avalanche-node-manually.md
+++ b/docs/nodes/build/run-avalanche-node-manually.md
@@ -108,19 +108,17 @@ Download the AvalancheGo repository into your `$GOPATH`:
 cd $GOPATH
 mkdir -p src/github.com/ava-labs
 cd src/github.com/ava-labs
-git clone git@github.com:ava-labs/avalanchego.git
+git clone https://github.com/ava-labs/avalanchego.git
 cd avalanchego
 ```
 
 :::info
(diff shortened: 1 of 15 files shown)


docs/nodes/build/run-avalanche-node-manually.md
```diff
@@ -108,19 +108,17 @@ Download the AvalancheGo repository into your `$GOPATH`:
 cd $GOPATH
 mkdir -p src/github.com/ava-labs
 cd src/github.com/ava-labs
-git clone git@github.com:ava-labs/avalanchego.git
+git clone https://github.com/ava-labs/avalanchego.git
```
yulin-dong, line 111: If we change this, we should make consistent in all docs to use https.

laviniat1996: I did check before opening the PR, and it does look like https is used everywhere in the docs (at least it will be after merging this PR).

dubrado (dismissed):

docs/nodes/build/run-avalanche-node-manually.md
laviniat1996: Actually, I've missed one place. Fixed the mistake now. HTTPS is used everywhere, and next to it there is an info box about SSH.

meaghanfitzgerald (approved):
lgtm :)

PR #1389, opened 2023-06-07
Add Explanatory Fuji

meaghanfitzgerald:
## Why this should be merged

Our docs lack a high-level overview of Fuji. 

## How this works

This doc defines Fuji as our official testnet, briefly defines what a faucet is, and explains the value the testnet provides a user/developer. Fuji and The Primary Network were added to a "Networks" subsection which is expanded by default. 

## Workflow checks

- [x] ran `vale /docs/file/path` on all changed `.md` files to ensure all grammar rules pass
- [x] ran `markdownlint /docs/file/path` on all changed `.md` files to ensure all linting rules pass
- [x] completed the above two checks and all additional rules outlined in `style-checker-notes.md` to
  ensure all checks pass
- [x] all files which were moved from their current directory to a new path have had their paths
  redirected via the `_redirects` file
- [x] `sidebars.json` also reflects any changes mentioned in the previous bullet

### DIFF
--- a/docs/learn/avalanche/avalanche-platform.md
+++ b/docs/learn/avalanche/avalanche-platform.md
@@ -15,6 +15,10 @@ The Primary Network is a special [Subnet](subnets-overview.md) that runs three b
 - The Platform Chain [(P-Chain)](avalanche-platform.md#p-chain)
 - The Exchange Chain [(X-Chain)](avalanche-platform.md#x-chain)
 
+:::note
+Avalanche Mainnet is comprised of the Primary Network and all deployed Subnets.
+:::
+
 A node can become a validator for the Primary Network by staking at least **2,000 AVAX**.
 
 ![Primary network](/img/primary-network.png)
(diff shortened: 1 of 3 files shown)


laviniat1996:
LGTM. Just a suggestion, what if you add info about the API endpoint used for Fuji? https://api.avax-test.network/. It is listed here https://docs.avax.network/apis/avalanchego/public-api-server, but maybe it would be useful to add it.

meaghanfitzgerald:
> LGTM. Just a suggestion, what if you add info about the API endpoint used for Fuji? https://api.avax-test.network/. It is listed here https://docs.avax.network/apis/avalanchego/public-api-server, but maybe it would be useful to add it.

I did consider this, and decided against it because I would like to keep reference information separate from explanation. In the future this reference will be more apparent than it is now.

Andyvargtz:
One suggestion is that for someone that doesn't know absolutely anything about Avalanche, might worth to have an explicit explanation that the Fuji Testnet has its own primary network formed by X, P an C chains and its own Fuji Deployed Subnets. Specially because new comers tend to think "primary network = mainnet" which is not. Also the 1st section only makes refrence to test SC and applications, which might lead to think it is only for C-chain

usmaneth (dismissed):

Andyvargtz (approved):

Andyvargtz (approved):

PR #1422, opened 2023-06-21
Docs: Genesis file header info 

Andyvargtz:
# Docs PR Template

## Why this should be merged

Completes description of the genesis file, which now include details of the header fields

## How this works

Provide details on the Genesis file header fields, also updated the genesis gas limit parameter in header to match the hex value of 15,000,000 in fee config section.

Following the workflow checks

## Workflow checks

- [X] ran `vale /docs/file/path` on all changed `.md` files to ensure all grammar rules pass
- [X] ran `markdownlint /docs/file/path` on all changed `.md` files to ensure all linting rules pass
- [X] completed the above two checks and all additional rules outlined in `style-checker-notes.md` to
  ensure all checks pass
- [] if a doc was removed/deleted, the path to that doc must be redirected to a valid URL via the
  `_redirects` file
- [] all files that were moved from their current directory to a new path have had their paths
  redirected via the `_redirects` file
- [] `_redirects` were manually verified with the cloudflare preview link
- [] `sidebars.json` reflects all changes made to file path

### DIFF
--- a/docs/subnets/customize-a-subnet.md
+++ b/docs/subnets/customize-a-subnet.md
@@ -196,7 +196,8 @@ hexadecimal and is equal to `15,000,000`.
 
 `coinbase`
 
-Refers to the address of the block producers. This also means it represents the recipient of the block reward. It is usually set
+Refers to the address of the block producers. This also means it represents the recipient of the 
+block reward. It is usually set
 to `0x0000000000000000000000000000000000000000` for the genesis block. To allow fee recipients in 
 Subnet-EVM, refer to [this section.](#setting-a-custom-fee-recipient)
 
--- a/docs/subnets/customize-a-subnet.md
+++ b/docs/subnets/customize-a-subnet.md
@@ -196,8 +196,7 @@ hexadecimal and is equal to `15,000,000`.
 
 `coinbase`
 
-Refers to the address of the recipient of the block reward. This is included in the first 
-transaction of each block of the blockchain. It is usually set
+Refers to the address of the block producers. This also means it represents the recipient of the block reward. It is usually set
 to `0x0000000000000000000000000000000000000000` for the genesis block. To allow fee recipients in 
 Subnet-EVM, refer to [this section.](#setting-a-custom-fee-recipient)
 
--- a/docs/subnets/customize-a-subnet.md
+++ b/docs/subnets/customize-a-subnet.md
@@ -180,9 +180,7 @@ cryptographically mined, thus, from this aspect, is valid. Default value is `0x0
 
 The difficulty level applied during the nonce discovering process of this block. Default value is `0x0`.
 
-`timestamp`
-
-The timestamp of the creation of the genesis block. This is commonly set to `0x0`.
+`timestamp`: The timestamp of the creation of the genesis block. This is commonly set to `0x0`.
 
 `extraData`
 
--- a/docs/subnets/customize-a-subnet.md
+++ b/docs/subnets/customize-a-subnet.md
@@ -161,7 +161,7 @@ The fields `nonce`, `timestamp`, `extraData`, `gasLimit`, `difficulty`, `mixHash
 set to match the `gasLimit` set in the `feeConfig`. You do not need to change any of the other genesis
 header fields.
 
-#### `nonce`, `mixHash` and `difficulty`
+`nonce`, `mixHash` and `difficulty`
 
 These are remnant parameters from Proof of Work systems.
 For Avalanche, these don't play any relevant role, so you should just leave them as: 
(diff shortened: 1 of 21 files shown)


docs/subnets/customize-a-subnet.md
```diff
@@ -152,6 +161,53 @@ The fields `nonce`, `timestamp`, `extraData`, `gasLimit`, `difficulty`, `mixHash
...
 header fields.
 
+`nonce`, `mixHash` and `difficulty`
+
+These are remnant parameters from Proof of Work systems.
+For Avalanche, these don't play any relevant role, so you should just leave them as: 
+
+- `nonce` as `0x0`.
+- `mixHash` as `0x0000000000000000000000000000000000000000000000000000000000000000`.
+- `diffuculty` as `0x0.
```
ceyonur, line 171:
```suggestion
- `difficulty` as `0x0`.
```

docs/subnets/customize-a-subnet.md
```diff
@@ -152,6 +161,53 @@ The fields `nonce`, `timestamp`, `extraData`, `gasLimit`, `difficulty`, `mixHash
...
+The total amount of gas that can be used in a single block. It should be set to
+the same value as in the [fee config](#fee-config). The value `e4e1c0` is
+hexadecimal and is equal to `15,000,000`.
+
+`coinbase`
+
+Refers to a special transaction included in each block of the blockchain. It is the first 
+transaction within the block and serves as the mechanism through which the block 
+reward is generated. It is usually set
+to `0x0000000000000000000000000000000000000000` for the genesis block.
```
ceyonur, line 192: This sounds like `coinbase` itself is the transaction. `coinbase` is an recipient address for block rewards (in our case only tx fees) that could be collected by block producers. In other words this is the address of whomever produced the block (i.e miner)

docs/subnets/customize-a-subnet.md
```diff
@@ -152,6 +161,53 @@ The fields `nonce`, `timestamp`, `extraData`, `gasLimit`, `difficulty`, `mixHash
...
+`gasUsed`
+
+This is the amount of gas used by the genesis block. It is usually set to `0x0`.
+
+`number`
+
+This is the number of the genesis block. It is usually set to `0x0`.
+
+
+
```
ceyonur, line 210: extra whitespaces here

meaghanfitzgerald:
Other than the nits above lgtm

docs/subnets/customize-a-subnet.md
ceyonur, line 171: Why we skipped explaining these? I know they're not very significant but again none of these Header fields actually matter and they should be left as they're (in most cases). So I feel like if other header fields need explanation, these 3 fields also should be explained too.

docs/subnets/customize-a-subnet.md
```diff
@@ -152,6 +161,53 @@ The fields `nonce`, `timestamp`, `extraData`, `gasLimit`, `difficulty`, `mixHash
...
+`0x0000000000000000000000000000000000000000000000000000000000000000` for the
+genesis block.
+
+`gasUsed`
+
+This is the amount of gas used by the genesis block. It is usually set to `0x0`.
+
+`number`
+
+This is the number of the genesis block. It is usually set to `0x0`.
```
ceyonur, line 207: This required to be `0x0` for the genesis. Otherwise it will error.

docs/subnets/customize-a-subnet.md
Andyvargtz: You're right! it does sound like that. For Avalanche, I understand there is no a single "miner" for each block, so for every block other than genesis, how does this works? what address its registered there?

ceyonur: Blocks are still proposed/built by single entities (miner/producer/proposer etc). 

The block structure of Coreth and Subnet-EVM are very similar to Eth blocks. Coinbase is always a part of coreth and subnet-evm blocks and represent the rewards address of whomever created the block. In Coreth this address has to be a blackhole address meaning that we burn all fees. However in Subnet-EVM there is an optional config (`allowFeeRecipients`) that can re-enable distributing these fees to the block producers through the coinbase. So in Subnet-EVM the coinbase can still represent block producer addresses.

docs/subnets/customize-a-subnet.md
```diff
@@ -196,10 +196,10 @@ hexadecimal and is equal to `15,000,000`.
 
 `coinbase`
 
-Refers to a special transaction included in each block of the blockchain. It is the first 
-transaction within the block and serves as the mechanism through which the block 
-reward is generated. It is usually set
-to `0x0000000000000000000000000000000000000000` for the genesis block.
+Refers to the address of the recipient of the block reward. This is included in the first 
+transaction of each block of the blockchain. It is usually set
```
ceyonur, line 200:
```suggestion
Refers to the address of the block producers. This also means it represents the recipient of the block reward. It is usually set
```

docs/subnets/customize-a-subnet.md
```diff
@@ -97,11 +97,16 @@ times. Changing these may cause issues, so treat them carefully.
 
 #### Fee Config
 
-`gasLimit`: Sets the max amount of gas consumed per block.
+`gasLimit`: Sets the max amount of gas consumed per block. Keep in mind that this will limit how 
+much computation can be processed in a single block and therefore will set an upper limit for 
+the maximum amount of gas that a single transaction can use.
```
meaghanfitzgerald, line 102:
```suggestion
`gasLimit`: Sets the max amount of gas consumed per block. This restriction puts a cap on the
amount of computation that can be done in a single block, which in turn sets a limit on the 
maximum gas usage allowed for a single transaction.
```

docs/subnets/customize-a-subnet.md
```diff
@@ -97,11 +97,16 @@ times. Changing these may cause issues, so treat them carefully.
...
+For reference, C-Chain value is set to `15,000,000`.
 
 `targetBlockRate`: Sets the target rate of block production in seconds. A target of 2 will target
-producing a block every 2 seconds. If the network starts producing faster than this, base fees are
+producing a block every 2 seconds. If the network starts producing faster than this, means that 
+more blocks than expected are being issued to the network, so the base fees will
 increased accordingly.
```
meaghanfitzgerald, line 108:
```suggestion
producing a block every 2 seconds. If the network starts producing blocks at a faster rate, it 
indicates that more blocks than anticipated are being issued to the network, resulting in an 
increase in base fees.
```

docs/subnets/customize-a-subnet.md
```diff
@@ -152,6 +161,62 @@ The fields `nonce`, `timestamp`, `extraData`, `gasLimit`, `difficulty`, `mixHash
...
+The combination of `nonce` and `mixHash` allows to verify that the Block has really been 
+cryptographically mined, thus, from this aspect, is valid. Default value is `0x0000000000000000000000000000000000000000000000000000000000000000`.
+
+`difficulty`
+
+The difficulty level applied during the nonce discovering process of this block. Default value is `0x0`.
+
+`timestamp`
+
+The timestamp of the creation of the genesis block. This is commonly set to `0x0`.
```
meaghanfitzgerald, line 185: Please adjust all param definitions to match throughout the doc. For example:

```suggestion
`timestamp`: The timestamp of the creation of the genesis block. This is commonly set to `0x0`.
```

meaghanfitzgerald (approved):

PR #1469, opened 2023-07-28
docs: Add note on VM binary naming

exdx:
This PR adds a small note detailing the fact that the VM binary name should not be manually altered.

Signed-off-by: Dan Sover <dan.sover@avalabs.org>


meaghanfitzgerald:
opened new req https://github.com/ava-labs/avalanche-docs/pull/1474

PR #1516, opened 2023-09-13
add gtm for GA4 and cookies banner

szmurlo:
migrating the GA4 tag to a GTM container, which also has Termly compliance integrated.


meaghanfitzgerald (approved):
lgtm

PR #1547, opened 2023-10-10
update: avalanchego-installer.sh

ashucoder9:
## Why this should be merged

The Installer script is not supported for macOS. But it doesn't throw an error message indicating that the user is on an unsupported OS. Instead, it simply exits after the `Preparing environment...` line.

Made changes to check if the user is running on macOS and terminated with a self-explanatory message.

## How this works

The problem is that the `getOsType` function is not compatible with macOS. It simply doesn't do anything. Not even a blank space or garbage value is returned. This results in the `osType` check failing and the program halts without any error message.

Effectively, [this](https://github.com/ava-labs/avalanche-docs/blob/master/scripts/avalanchego-installer.sh#L244) message should echo on the terminal.

I have included commands in the `getOsType` function to check if the OS is different than Linux. This was initially being checked [here](https://github.com/ava-labs/avalanche-docs/blob/master/scripts/avalanchego-installer.sh#L250).

Ran the script again to test and the below image is the result.

Screenshot 2023-10-10 at 3 20 21 PM

### DIFF
--- a/scripts/avalanchego-installer.sh
+++ b/scripts/avalanchego-installer.sh
@@ -132,9 +132,15 @@ check_reqs_rhel () {
   fi
 }
 getOsType () {
+foundOS="$(uname)"                              #get OS
+if [ "$foundOS" = "Linux" ]; then
   which yum 1>/dev/null 2>&1 && { echo "RHEL"; return; }
   which zypper 1>/dev/null 2>&1 && { echo "openSUSE"; return; }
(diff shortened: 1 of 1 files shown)


meaghanfitzgerald:
LGTM, would love some feedback from @jpop32 before merging. Thanks as always @ashucoder9 💪

meaghanfitzgerald (approved):

PR #1574, opened 2023-11-02
Deprecated IPC API

meaghanfitzgerald:
## Why this should be merged

The IPC API has been deprecated: https://github.com/ava-labs/avalanchego/pull/2168#issue-1942101224

## How this works

removed all relevant config flags


joshua-kim (approved):

meaghanfitzgerald:
Noticed IPC flags are still located in `config/flags.go` https://github.com/ava-labs/avalanchego/blob/20f3580b4912f29adc602ad883e84897f0965cc1/config/flags.go#L322 . 

I noticed the above while I was looking for a command to print all flags with definitions so that we can reconcile the other deprecated/accepted flags in the docs. Is there is an easy way to do this? @joshua-kim

joshua-kim:
> Noticed IPC flags are still located in config/flags.go https://github.com/ava-labs/avalanchego/blob/20f3580b4912f29adc602ad883e84897f0965cc1/config/flags.go#L322 .
>
> I noticed the above while I was looking for a command to print all flags with definitions so that we can reconcile the other deprecated/accepted flags in the docs. Is there is an easy way to do this? @joshua-kim

You can run `--help` when you run avalanchego to print out all of the flags but this won't include the ones that are deprecated (e.g ipcs).

PR #1582, opened 2023-11-16
Fix unattended avalanchego installer flow

jpop32:
# Docs PR Template

## Why this should be merged

Fixes unattended installer flow so it doesn't require input if the public rpc option is provided as an argument.

## How this works

Checks for provided rpc flag and skips interactive rpc setup if found.

Tested locally on my node.


ashucoder9 (approved):

meaghanfitzgerald (approved):
LGTM thanks!

Issue #1592, opened 2023-11-27
Use exact versioning in package.json


yulin-dong:
**Describe the bug**
Exact versions should be used in package.json to avoid potential bugs by not-exact version upgrade of dependencies.

laviniat1996:
Hi @yulin-dong , can you provide an example for exact versioning?

yulin-dong:
https://github.com/ava-labs/avalanchejs/blob/master/package.json uses exact version, instead of ^

laviniat1996:
@yulin-dong Sorry for the delay. I'll look into it this week.

laviniat1996:
Thank you @meaghanfitzgerald

PR #1614, opened 2023-12-13
Package management, admonition correction and consensus page changes

meaghanfitzgerald:
## Why this should be merged

Continuous docs improvements

## How this works

- Makes all packages in `package.json` absolute, addresses https://github.com/ava-labs/avalanche-docs/issues/1592
- Removes consensus comparison image from the consensus page
- corrects 2 admonitions that were incorrectly translated by the translation bot


i18n/es/docusaurus-plugin-content-docs/current/nodes/validate/add-a-validator.md
```diff
@@ -20,7 +20,7 @@ qué blockchains. Para agregar un validador, emitiremos
 [transacciones](http://support.avalabs.org/en/articles/4587384-what-is-a-transaction)
 a la P-Chain.
 
-:::advertencia
+:::warning
```
meaghanfitzgerald, line 24: @laviniat1996 when using the translation bot always double check to make sure the [admonitions ](https://docusaurus.io/docs/markdown-features/admonitions) are in english

laviniat1996 (approved):

PR #1664, opened 2024-02-08
Update `getNodeID` to include BLS info

laviniat1996:


meaghanfitzgerald:
font in gif looks very small 
image

why does this endpoint need a gif? the others do not have and making these api calls should be self-explanatory

laviniat1996:
@meaghanfitzgerald The BLS signature and BLS key are needed now for adding validators, thus very important information. Kieran asked me to create a video/gif to showcase how to get them.

meaghanfitzgerald:
The POP values are clearly outlined in the example response. Making a curl request and reading the JSON response is a very common activity for developers across all industries. Kieran's request should be handled with an adhoc video or a support article, and does not need to be outlined in our docs.

meaghanfitzgerald (approved):

PR #1668, opened 2024-02-12
feat(AvalancheGo): add ACP Signaling flags in docs

ashucoder9:
## Why this should be merged

ACP Signaling flags make it easy for validator node operators to participate in the ACP governance. We need appropriate reference docs for the config flags in order to make sure it's easy to follow.

Checked on my local node.

## Additions needed

@joshua-kim we'd like to add info on how a user can specify multiple ACPs in one single request. will also need info on if and how a user can change his opinion from support to conflict or vice-versa on an ACP


meaghanfitzgerald (dismissed):
lgtm. would ignore the vale errors

meaghanfitzgerald (approved):

PR #1658, opened 2024-02-07
add doc for getSubnet

felipemadero:
# Docs PR Template

## Why this should be merged
Documents an upcoming API call. getSubnets. ref https://github.com/ava-labs/avalanchego/pull/2704

## How this works
add entry into pchain api doc

## How these changes were tested 
they where copied from the console output of a method execution

## To prevent any errors while building

- [] run `vale /docs/file/path` on all changed `.md` files to ensure all grammar rules pass
- [] run `markdownlint /docs/file/path` on all changed `.md` files to ensure all linting rules pass
- [] complete the above two checks and all additional rules outlined in `style-checker-notes.md` to
  ensure all checks pass

### If a document was removed/deleted

- [] the path to that doc must be redirected to a valid URL via the
  `_redirects` file
- [] `_redirects` were manually verified with the cloudflare preview link
- [] `sidebars.json` reflects all changes made to file path

### If a document was moved

- [] all files that were moved from their current directory to a new path have had their paths
  redirected via the `_redirects` file
- [] `_redirects` were manually verified with the cloudflare preview link
- [] `sidebars.json` reflects all changes made to file path

PR #1673, opened 2024-02-14
Revert "add doc for getSubnet"

meaghanfitzgerald:
Reverts ava-labs/avalanche-docs#1658

API method is not in production yet


meaghanfitzgerald:
@felipemadero

PR #1688, opened 2024-02-22
Add custom admonitions in Teleporter remote docs to drive users to Academy

ashucoder9:
## Goal

We want to add an admonition box on each Teleporter docs which directs users to the Academy course.

## How it was done

Since the Teleporter docs are fetched from remote files, we don't have much control over their content. To put the admonition box after the first line of each doc (which is their title as well), I created the `insertLinesAfterFirstLine` method, which takes the content and adds the admonition lines to the document.

cc @martineckardt


meaghanfitzgerald (approved):
nice!

PR #1696, opened 2024-03-01
Durango updates + Add validator tutorial updates

laviniat1996:


docs/nodes/validate/how-to-stake.md
```diff
@@ -78,6 +78,7 @@ probability that a given validator is sampled is proportional to its stake.
 When you add a node to the validator set, you specify:
 
 - Your node’s ID
+- Your node's BLS key and BLS signature
 - When you want to start and stop validating
```
felipemadero, line 82: you can't specify when to start

docs/nodes/validate/how-to-stake.md
```diff
@@ -78,6 +78,7 @@ probability that a given validator is sampled is proportional to its stake.
...
 - How many AVAX you are staking
```
felipemadero, line 83: you can also add a comment on this file, regarding backing up the bls key file, on `Secret Management` section

docs/nodes/validate/how-to-stake.md
felipemadero: also, you don't specify when delegator starts (`Delegators` section)

felipemadero: also, under `Error: Couldn't Issue TX: Validator Would Be Over Delegated` section, there is no starttime check anymore

docs/nodes/validate/add-a-validator.md
```diff
@@ -141,166 +143,94 @@ to run the example scripts directly from an AvalancheJS directory.
...
+Now let's say you want the staking period to end on a specific 
+date and time, for example May 15, 2024, at 11:20 AM.
+It can be achieved as shown in the code below.
 
 ```ts
-const startTime: BN = new BN(1654656829); // Wed Jun 08 2022 02:53:49 GMT+0000
-const endTime: BN = new BN(1662602029); // Thu Sep 08 2022 01:53:49 GMT+0000
+const startTime = await new PVMApi().getTimestamp();
+const startDate = new Date(startTime.timestamp);
+const start = BigInt(startDate.getTime() / 1000);
```
felipemadero, line 229: maybe you want to clarify that the api call requires a start argument for keeping back compatibility, but it will be ignored

PR #1754, opened 2024-05-01
change 0 timestamps to current timestamp

ceyonur:
# Docs PR Template

## Why this should be merged

Durango upgrade are activated at the same time with Avalanche network. Providing a 0 timestamp to Durango means it will default to a suitable time for the network (depends on the network id a.k.a Fuji, Mainnet etc). However providing a 0 timestamp in genesis block would not activate Durango right-away, instead it will activate the Durango in the first block (genesis considered as the 0th block). This could cause some issued in gas estimations, calling and deploying contracts in fresh chains. Also 0 timestamps in genesis known to create issues with ProposerVM.

Instead of providing `0x0`=0 timestamp, we should recommend creating subnet chains with a recent timestamp that would naturally activate Durango (and all previous) upgrades.

Tested in Local networks + rolled out to the test networks.


docs/build/subnet/deploy/custom-vm-subnet.md
```diff
@@ -113,7 +103,16 @@ compatible with your custom VM.
...
 ```
 
-Open a text editor and copy the preceding text into a file called `custom_genesis.json`.
+Open a text editor and copy the preceding text into a file called `custom_genesis.json`. For full breakdown of the genesis file, see the
+[Genesis File](/build/subnet/upgrade/customize-a-subnet.md#genesis).
+
+:::note
+
+The `timestamp` field is the Unix timestamp of the genesis block. `0x66321C34` represents
+the timestamp 1714560052 which is the time this tutorial was written. You can use the
```
darioush, line 112:
```suggestion
the timestamp 1714560052 which is the time this tutorial was written. You should use the
```

docs/build/subnet/upgrade/customize-a-subnet.md
```diff
@@ -94,7 +92,7 @@ You can use `eth_getChainConfig` RPC call to get the current chain config. See
 
 #### Hard Forks
 
-`homesteadBlock`, `eip150Block`, `eip150Hash`, `eip155Block`, `byzantiumBlock`, `constantinopleBlock`,
+`homesteadBlock`, `eip150Block`, `eip155Block`, `byzantiumBlock`, `constantinopleBlock`,
 `petersburgBlock`, `istanbulBlock`, `muirGlacierBlock` are EVM hard fork activation
```
darioush, line 96: should we add the other eth forks?

darioush (dismissed):

darioush (dismissed):

laviniat1996 (approved):

PR #1861, opened 2024-08-29
add: search index for integration pages

ashucoder9:
till now, the search route only creates index for the Docs pages. this PR configure the search route file and integration-loader to also create index for integration pages.

TODO: replicate this on Academy (to include guides in search)


app/api/search/route.ts
```diff
@@ -1,11 +1,20 @@
 import { getPages } from '@/utils/docs-loader';
```
martineckardt, line 1: Let's refactor the docs-loader to export { getDocsPage, getDocsPages, docsPageTree }

martineckardt (changes requested):

martineckardt (approved):

PR #1876, opened 2024-09-06
add: script for fetching remote github content

ashucoder9:
- added axios and tsx to fetch content and execute script during build
- added remote-content.mts file where all the logic to fetch and display content sits

this won't render properly on preview until we merge and update the custom build command on Vercel.


ashucoder9:
cc @meaghanfitzgerald API docs are not added yet, only cross-chain docs

meaghanfitzgerald:
`yarn.lock` should be added to the gitignore and the file removed from the repo. otherwise lgtm 👍

utils/remote-content.mts
```diff
@@ -0,0 +1,143 @@
+import axios from 'axios';
+import * as fs from 'fs';
+import * as path from 'path';
+
+interface FileConfig {
```
meaghanfitzgerald, line 5: hell yes

meaghanfitzgerald (dismissed):

meaghanfitzgerald (approved):

PR #1885, opened 2024-09-12
Subnet 9000 Replacement

ashucoder9:
- [x] Replace Subnet with Avalanche L1 in content without replacing commands
- [x] Modify files and folder names
- [x] Reflect file name changes in the link structure across modified pages
- [x] Make sure content like Subnet-EVM is untouched while replacement
- [ ] Make sure the context of "Subnet = Set of Validators" is maintained throughout the docs
- [ ] Modify redirects to reflect the latest changes

PS: manually update the words wherever a better context is required while making the changes.

also, the remote docs are not updated yet (specifically cross-chain folder atm).


content/docs/api-reference/p-chain/api.mdx
```diff
@@ -1950,7 +1950,7 @@ curl -X POST --data '{
 
 ### `platform.validatedBy`
 
-Get the Subnet that validates a given blockchain.
+Get the Avalanche L1 that validates a given blockchain.
```
ashucoder9, line 1953: @meaghanfitzgerald what's the recommend way to tackle such contexts? shall we use Subnet or instead a "set of validators"? do you think we have a specific word for subset of validators or subnetID?

iJaack: @ashucoder9 from the outside, I believe the term Subnet was double-used before. A subnet is a subset of validators, se that term was correct before Avalanche9000. Given that now validators don’t need to validate the Primary Network, it could be tricky to call it a ‘subset’, since it’s not technically a subset. But, they still need to connect to the P-Chain in some way, so if we change the description of the Primary Network to ‘Primary Subnet = the Subnet validating the Primary Chains’ or something similar, then it would make sense to still call the Subnet, Subnet.

So the writing would be ‘Get the Subnet that validates a given Avalanche L1’.

meaghanfitzgerald: Subnet has always meant _a set of validators that validate a given chain or set of chains_. In the context of the Avalanche Network pre-Etna, the Subnet validators were a subset of Avalanche's Network Validators. Post-etna, this will no longer be a given. 

L1 is effectively interchangeable with Subnet.
- Subnet validators -> L1 validators
- Subnet chain -> L1 chain/ L1 blockchain
- etc. 

The change made above is valid, but a bit more context will be helpful for new devs:

```suggestion
Get the `SubnetID` of the Avalanche L1 that validates a given blockchain.
```

`SubnetID` will be renamed in the future, but in order to prevent delaying the upgrade and downstream breaking changes, this value will keep it's name for now.

owenwahlgren (approved):
LGTM!

content/docs/api-reference/p-chain/api.mdx
ashucoder9:
```suggestion
Get the `SubnetID` of the Avalanche L1 that validates a given blockchain.
```

ashucoder9:
@owenwahlgren can you approve the PR and merge?

owenwahlgren (approved):
Looks good to me 🫡

PR #1979, opened 2024-12-20
remove eth_getAssetBalance

darioush:
ANT (avalanche native tokens) precompiles have been disabled completely since ApricotPhase 6, and there is no intention to re-enable them in the future.
We plan to remove these from coreth: https://github.com/ava-labs/coreth/pull/715.


meaghanfitzgerald (approved):
TYVM for raising this PR, approved

PR #1996, opened 2025-01-13
feat(api/c-chain): change of behavior for eth_getProof

qdm12:


content/docs/api-reference/c-chain/api.mdx
```diff
@@ -81,6 +81,12 @@ For batched requests on the [public API node](/tooling/rpc-providers) , the maxi
 number of items is 40. We are working on to support a larger batch size.
 </Callout>
 
+Note the `eth_getProof` is acting differently from release v0.14.1:
+
+- On archive nodes (`"pruning-enabled": false`): queries for historical proofs for state older than approximately 24 hours preceding the last accepted block will be rejected by default. This can be adjusted with the new option `historical-proof-query-window` which defines the number of blocks before the last accepted block which should be accepted for state proof queries, or set to `0` to accept any block number state query.
+- On `pruning` nodes: queries for proofs past the tip buffer (32 blocks) will be rejected
```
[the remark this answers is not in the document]
darioush: It's not the exact same behavior as they would have responded to proofs for blocks matching the commit height 4096, so we can leave this here.

Also perhaps this should be under the Ethereum APIs section, we can add a new section for `eth_getProof` if there's not one already and just include this notice, mentioning the API otherwise behaves as upstream.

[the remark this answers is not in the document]
qdm12: It's already in the `## Ethereum APIs` section / `### Standard Ethereum APIs` sub-section.
I've moved this to an `#### Exceptions` subsection as a bullet point, let me know if that works.

content/docs/api-reference/c-chain/api.mdx
```diff
@@ -81,6 +81,14 @@ For batched requests on the [public API node](/tooling/rpc-providers) , the maxi
...
 </Callout>
 
+#### Exceptions
+
+- `eth_getProof` behaves differently than geth, from release v0.14.1, with the following differences:
+  - On archive nodes (`"pruning-enabled": false`), queries for proofs for state older than approximately 24 hours preceeding the last accepted block will be rejected by default.
+    This can be adjusted with the option `historical-proof-query-window` which defines the number of blocks before the last accepted block which should be accepted for state proof queries, or set it to `0` to accept any block number state query.
+  - On pruning nodes (`"pruning-enabled": true`), queries for proofs past the tip buffer of 32 blocks are always rejected
+
+
```
meaghanfitzgerald, line 91:
```suggestion
`eth_getProof` behaves differently than geth, from release `v0.14.1`, with the following differences:
- On archival nodes (nodes with`pruning-enabled` set to `false`), queries for state proofs older than 24 hours preceding the last accepted block will be rejected by default. This can be adjusted with `historical-proof-query-window`, which defines the number of blocks before the last accepted block that can be queried for state proofs. Set this option to `0` to accept a state query for any block number.  
- On pruned nodes (nodes with `pruning-enabled` set to `true`), queries for proofs past a 32 block buffer are always rejected.

```

meaghanfitzgerald (commented):
suggestions to improve readability. also removes extra blank line to please the .md linter

content/docs/api-reference/c-chain/api.mdx
qdm12: 1. TIL I've been using `preceeding` wrong forever, thanks! Only 1 `e`!
1. Isn't it `pruning nodes` instead of `pruned nodes`? We're talking about avalanche nodes, not i.e. trie nodes 🤔 
1. `queries for state proofs past a 32 block buffer are always rejected.` ➡️  `queries for state proofs outside the 32 blocks window trailing the last accepted block are always rejected.` - since the `32 block buffer` might leave some room for interpretation and is a bit of an implementation details I'd think?

Thanks!

content/docs/api-reference/c-chain/api.mdx
```diff
@@ -81,6 +81,14 @@ For batched requests on the [public API node](/tooling/rpc-providers) , the maxi
 number of items is 40. We are working on to support a larger batch size.
 </Callout>
 
+#### Exceptions
+
+- `eth_getProof` behaves differently than geth, from release v0.14.1, with the following differences:
```
qdm12, line 86: the release version should the avalanchego version, not the coreth version. I suppose we should wait for the next avalanchego release using coreth v0.14.1 (unreleased) to merge this PR.

content/docs/api-reference/c-chain/api.mdx
meaghanfitzgerald: 1. 🫡 
2. 🫡 
3. corrected
4. reworded a bit differently

meaghanfitzgerald:
I am surprised this doesn't need a rebasing

content/docs/api-reference/c-chain/api.mdx
meaghanfitzgerald: please "sign off and commit merge" if the changes are good to go

content/docs/api-reference/c-chain/api.mdx
```diff
@@ -81,6 +81,12 @@ For batched requests on the [public API node](/tooling/rpc-providers) , the maxi
 number of items is 40. We are working on to support a larger batch size.
 </Callout>
 
+#### Exceptions
+
+`eth_getProof` behaves differently than geth, from release `v0.14.1`, with the following differences:
```
qdm12, line 86:
```suggestion
`eth_getProof` behaves differently than geth, from release [`v0.12.2`](https://github.com/ava-labs/avalanchego/releases/tag/v1.12.2), with the following differences:
```

PR #2018, opened 2025-01-23
hide: avagpt button on small screens

ashucoder9:
it's an obstacle while navigating b/w docs + switching dark/light mode from navbar


owenwahlgren (approved):

owenwahlgren (approved):
LGTM!

PR #2024, opened 2025-01-28
PoA Validator Manager Web UI

owenwahlgren:
Manage your PoA validator set entirely from frontend, only with signature aggregator as local dependency.

- Add Validator [✅]
- Remove Validator [❌]
- Change Weight [❌]


martineckardt:
- Let's call this tool PoA Validator Management
- URL: /tools/poa-validator-management

martineckardt (approved):

PR #2030, opened 2025-01-30
Proxy in Genesis

containerman17:
One problem - can not initialize PoAValidatorManager due to lost warp message even with Core. WIP


martineckardt:
LGTM

PR #2082, opened 2025-02-12
Add infraBuidl(AI) and Avalanche Innovation House in Grants, Add Banner

federiconardelli7:
Add Banner for Innovation House in Home and Grants.


ashucoder9:
the preview is not correctly rendered. there is a margin introduced in the navbar and banner should be visible on all pages where navbar is present (unless user closes), please add close option too

martineckardt (approved):

PR #2125, opened 2025-02-25
chore: fix JSON syntax issue in create_config_file

letshard:
Found and fixed a bug in `create_config_file` that could produce invalid JSON.  

Previously, the `commaAdd` variable was appended even to the last element, causing a syntax error. Now, if `state-sync-enabled` is the last item, the comma is no longer added.  

The generated file is now valid.


felipemadero (approved):
LGTM

PR #2150, opened 2025-03-03
Minor fixes for typos, formatting, and obsolete information within multi-chain architecture module

gonzales17:


martineckardt (approved):
Thanks a lot. Great catch

PR #2170, opened 2025-03-10
chore: escape $ char to cancel LaTeX formatting

LMulvey:
Screenshot 2025-03-10 at 1 04 40 PM

Currently, the two `$` characters are causing this string of text to be treated as LaTeX. This should correct that issue.


martineckardt (approved):

martineckardt:
Thank you 🙌🏻

PR #2181, opened 2025-03-13
new: Fumadocs v15 upgade

ashucoder9:


ashucoder9:
this is missing the integrations landing page so far (everything else looks good and works). adding that page causes some random node:url issues. will release a fix.

martineckardt (approved):

PR #2197, opened 2025-03-17
add: rating feature for docs and academy

ashucoder9:


martineckardt:
What do you think about combining it with the edit and report issue buttons?

image

martineckardt (approved):

PR #2200, opened 2025-03-18
unit converter and cipher converter

alejandro99so:


martineckardt:
- [ ] Remove requirement to connect wallet

PR #2245, opened 2025-03-25
RPC http safeguard check

navillanueva:
Created component to notify the user when he would be making a call to an http endpoint (as the toolbox is secured by https) and give two options to either secure the http endpoint via a reverse proxy or how to run the toolbox locally


toolbox/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+## Installation
+
+1. Clone the repository:
+```bash
+git clone https://github.com/ava-labs/builders-hub.git
+cd builders-hub
+```
+
+2. Install dependencies:
```
martineckardt, line 13: The toolbox is a separate NPM package with its own dependencies. The user needs to cd into that directory first and then install dependencies there. 

Will prob also work this way because the builder hub has the toolbox, but will if the user just wants to run the toolbox a lot of unneeded dependencies will be downloaded

martineckardt (changes requested):

martineckardt (approved):

PR #2267, opened 2025-03-27
add glossary blockchain/avalanche key terms 

federiconardelli7:


martineckardt:
Let's chat about this in our next call

ashucoder9:
will start this afresh and get this into working state

PR #2276, opened 2025-03-29
Update system-requirements.mdx

UltravioletaDAO:
Because Ubuntu 20.04 is reaching the end of its support window, the binaries after v1.13.0 release do not support Ubuntu 20.04 (Focal).


martineckardt (approved):
Thank you 🙏🏻

PR #2312, opened 2025-04-11
added KYC/Identity Verification into integrations

federiconardelli7:
DO NOT Merge. Update is required.
- Delete Quadrata. 
- Add parallel markets
- Add keyring
- search and maybe add Sundial (no, added comment with explaination)
- Add Jumio


martineckardt (dismissed):

federiconardelli7:
Regarding Sundial, from the search results, it appears to be an early-stage project focused on ZK-based on-chain passport solutions. There isn't enough detailed information available

federiconardelli7:
ready to be merged.

PR #2350, opened 2025-04-21
create utility for builderhub og images generation code reusable

federiconardelli7:


federiconardelli7:
Checked and it works fine. Remove duplicative code, now a third parameter path is passed and that will be highlighted in red in the og image.

app/api/og/guides/[slug]/route.tsx
```diff
@@ -1,149 +1,19 @@
...
-);
-
-const regular = fetch(new URL('../../GeistMono-Light.ttf', import.meta.url)).then((res) =>
-  res.arrayBuffer(),
-);
-
-export async function GET(
-  request: NextRequest,
-): Promise<ImageResponse> {
+export async function GET(request: NextRequest) {
```
Comment (martineckardt) [inline app/api/og/guides/[slug]/route.tsx:6]: Why did we remove the type of the return value?

martineckardt (changes requested):

app/api/og/guides/[slug]/route.tsx
Comment (federiconardelli7) [reply app/api/og/guides/[slug]/route.tsx]: I think the typescript compiler should determine the return type from the async function createOGResponse in utils that owns the promise containing the type.
 export async function createOGResponse({
  title,
  description,
  path,
  fonts
}: OGProps & { fonts: { medium: ArrayBuffer, light: ArrayBuffer, regular: ArrayBuffer } }): Promise<ImageResponse> {
  return new ImageResponse(

I had tried it and it all worked correctly but I think it's better to make it explicit directly so that it's immediate to understand what the endpoint returns without having to look at the implementation itself. 

Now i pushed a commit with the explicit type return in the routes and checked again and everything seems to work fine.

martineckardt (approved):

PR #2419, opened 2025-05-07
Check precompiles activated

navillanueva:


navillanueva:
Here is a little preview of how it works changing between L1s with all of the precompiles deployed, and chains that dont have 

https://www.loom.com/share/cf472c9ebd884ead9a3dc9c6d851c9ea?sid=baae2b63-9423-42ab-9594-f03982ed69cb

It might be the case that the dispatch L1 does have precompiles deployed though, but if it was deployed through Avacloud the subnetEVM **getActiveRulesAt** query is b**locked by Avacloud at the moment**

It has been tested with a deployed L1 that has no precompiles (except the warp) if you want to test use this rpc url

https://3.90.156.8.nip.io/ext/bc/2u5zKRAJ3iMsGMMJQTz97okfZkW2tU6Un3LSeJZ7n9Gntdt2RR/rpc

toolbox/src/toolbox/Precompiles/DeployerAllowlist.tsx
```diff
@@ -1,20 +1,90 @@
...
+        setIsPrecompileActive(data?.result?.precompiles?.contractDeployerAllowListConfig?.timestamp !== undefined);
+      } catch (error) {
+        setIsPrecompileActive(false);
+      } finally {
+        setIsChecking(false);
+        setHasChecked(true);
+      }
+    }
+
+    checkPrecompileStatus();
```
navillanueva, line 67: put in its own component (compare to how the wallet component works) this is a common react pattern  (look into react patterns as well)

toolbox/src/toolbox/Precompiles/DeployerAllowlist.tsx
```diff
@@ -1,20 +1,90 @@
...
+
+    async function checkPrecompileStatus() {
+      try {
+        setIsChecking(true);
+        if (!viemChain) {
+          throw new Error('Chain not available');
+        }
+        const rpcUrl = viemChain.rpcUrls.default.http[0];
+
+        const response = await fetch(rpcUrl, {
```
navillanueva, line 40: refactor into a method in the coreviem

martineckardt (changes requested):

toolbox/src/coreViem/methods/getActiveRulesAt.ts
```diff
@@ -0,0 +1,33 @@
...
+            contractDeployerAllowListConfig?: { timestamp: number };
+            txAllowListConfig?: { timestamp: number };
+            feeManagerConfig?: { timestamp: number };
+            rewardManagerConfig?: { timestamp: number };
+            contractNativeMinterConfig?: { timestamp: number };
+        };
+    };
+}
+
+export async function getActiveRulesAt(rpcUrl: string): Promise<ActiveRulesResponse> {
```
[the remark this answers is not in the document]
navillanueva: Had a bit of a tough time with this, ultimately im not sure i achieved exactly what you wanted, i think the issue is the getEthereumChain is an api for the wallet, but the getActivesRulesAt ultimately i need the rpc to make the request

All in all i am passing the wallet client now, but i just use it to then get the chain information and then the rpc

toolbox/src/coreViem/methods/getActiveRulesAt.ts
```diff
@@ -0,0 +1,70 @@
...
+                IsFortuna: false,
+            },
+            precompiles: {}
+        };
+    }
+
+    // For other types of errors, throw
+    if (data.error) {
+        throw new Error(data.error.message || 'RPC call failed');
+    }
```
navillanueva, line 67: @martineckardt doubting about my error handling here, im not sure if dispatch was deployed with AvaCloud but i get back a "method doesn't exist" which is the same response i got back when testing with my Avacloud deployed L1. I think ultimately the user should just see "precompile not available" and thats what im covering here

toolbox/src/coreViem/methods/getActiveRulesAt.ts
```diff
@@ -0,0 +1,70 @@
+import { WalletClient } from "viem";
+import { CoreWalletRpcSchema } from "../rpcSchema";
+
+type GetActiveRulesAtResponse = Extract<CoreWalletRpcSchema[number], { Method: 'eth_getActiveRulesAt' }>['ReturnType'];
+
+export async function getActiveRulesAt(client: WalletClient<any, any, any, CoreWalletRpcSchema>): Promise<GetActiveRulesAtResponse> {
+    // Get the chain configuration from the wallet client
+    const chain = await client.request({
+        method: 'wallet_getEthereumChain',
```
martineckardt, line 9: Can you do eth_getActiveRules at here directly?

toolbox/src/coreViem/rpcSchema.ts
```diff
@@ -19,6 +19,38 @@ export type CoreWalletRpcSchema = [
             isTestnet: boolean
         }
     },
+    {
```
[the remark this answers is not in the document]
martineckardt: Yes, that's good. The component can extract whatever they need

toolbox/src/toolbox/Precompiles/DeployerAllowlist.tsx
```diff
@@ -1,20 +1,22 @@
...
-          precompileAddress={DEFAULT_DEPLOYER_ALLOWLIST_ADDRESS}
-          precompileType="Deployer"
-        />
-      </div>
-    </div>
+    <CheckPrecompile
+      configKey="contractDeployerAllowListConfig"
+      precompileName="Deployer Allowlist"
+    >
+      <AllowlistComponent
```
martineckardt, line 16: Much better

toolbox/src/coreViem/methods/getActiveRulesAt.ts
navillanueva: Conclusion - I had tested to call the method using the WalletClient but was getting a "Method doesn't exist" on the connected L1 where i had previously been able to use it 

Most likely core doesn't have this functionality available since it isn't a general method used or created by Ethereum

martineckardt (approved):

PR #2453, opened 2025-05-14
Academy icm adding toolbox

navillanueva:
**Key Changes**

- Created a new branch on avalanche-starter-kit that creates a codespace **without avalanche-cli or awm relayer**
- Removed creating a chain as part of the course
- Switched commands from using the deployed chain to echo/dispatch

**Codespaces Creation Times**
- _main_: 3 minutes
- _interchain-messaging_: 2 minutes 30 seconds
- _interchain-messaging_: (only foundry) 40 seconds 

**PR of new Avalanche-Starter-Kit** 
https://github.com/ava-labs/avalanche-starter-kit/pull/74


content/academy/interchain-messaging/03-avalanche-starter-kit/02-set-up.mdx
navillanueva, line 1: I tried turning the set-up.mdx into a component that took in a branch, but then the text stopped being processed as markdown

The complexity of trying to make a dynamic component work with MDX and Fumadocs would likely introduce more problems than it solves for what i gathered, but if there is a better way than copying everything to just change the branch it opens up the github codespaces with please let me know

content/academy/interchain-messaging/04-icm-basics/04-create-sender-contract.mdx
```diff
@@ -60,42 +60,18 @@ The key things to understand:
 
 ### Get destinationBlockchainID
 
-To get the blockchainID on our Avalanche L1 let's use the Avalanche-CLI built in feature with the command:
+To get the blockchainID of the Dispatch test L1, you can find it in the foundry.toml configuration file or use the following value:
```
martineckardt, line 63: We had this step because every L1 created has a different blockchainID. We could just hardcode the Dispatch blockchainID in the contracts in the starter kit and delete this step

content/academy/interchain-messaging/05-two-way-communication/05-create-the-receiver-contract.mdx
```diff
@@ -16,7 +16,7 @@ Go ahead and deploy the receiver contract:
 ### Deploy the Receiver Contract
 
 ```bash
-forge create --rpc-url myblockchain --private-key $PK contracts/interchain-messaging/send-roundtrip/receiverOnSubnet.sol:ReceiverOnSubnet --broadcast
+forge create --rpc-url dispatch --private-key $PK contracts/interchain-messaging/send-roundtrip/receiverOnSubnet.sol:ReceiverOnSubnet --broadcast
```
martineckardt, line 19: Should we call it dispatch-l1? Otherwise the students could think it's part of foundry command because the word dispatch is somewhat related to messaging.

Alternatively we could also do fuji-dispatch

Same for echo

content/academy/interchain-messaging/03-avalanche-starter-kit/02-set-up.mdx
```diff
@@ -1,13 +1,119 @@
...
+   - Open Core wallet
+   - Click your account name (top right)
+   - Select "View Private Key"
+   - Enter your password
+   - Copy your:
+     - Account address (0x...)
+     - Private key (0x...)
+
+<Callout type="warning">
+Never share your private key with anyone or commit it to version control. Keep it secure and backed up.
```
martineckardt, line 73: Add comment: do not use a private key holding real funds on testnet. Separate mainnet and testnet keys

content/academy/interchain-messaging/03-avalanche-starter-kit/02-set-up.mdx
```diff
@@ -1,13 +1,119 @@
...
+2. Add your Core wallet credentials to `.env`:
+```bash
+# Your private key for signing transactions (for testing only, never use in production)
+PK=your_private_key_here
+
+# Your funded address (the address derived from your private key)
+FUNDED_ADDRESS=your_funded_address_here
+
+# Teleporter and chain configuration 
+TELEPORTER_REGISTRY_FUJI_DISPATCH=your_dispatch_registry_address
```
martineckardt, line 95: We can pre-fill these

https://github.com/ava-labs/icm-contracts/tree/main/contracts/teleporter#deployed-addresses

ashucoder9:
@navillanueva academy course pages are returning a 500 (docs is not impacted, most probably due to the force-dynamic addition i believe)

content/academy/interchain-messaging/03-avalanche-starter-kit/05-foundry-quickstart.mdx
```diff
@@ -1,15 +1,76 @@
...
-import defaultMdxComponents from "fumadocs-ui/mdx";
+Deploy your contracts to either network using `forge create`:
 
-<FoundryQuickstart components={defaultMdxComponents}/>
\ No newline at end of file
+```bash
+# Deploy to Fuji C-Chain
+forge create --rpc-url fuji-c \
+  --private-key $PK \
+  src/contracts/YourContract.sol:YourContract \
```
ashucoder9, line 19: @navillanueva can we simplify this with a command which actually works and acts like an example? would be easier for people to follow

content/academy/interchain-messaging/03-avalanche-starter-kit/05-foundry-quickstart.mdx
```diff
@@ -1,15 +1,76 @@
...
+  --constructor-args $FUJI_C_CHAIN_BLOCKCHAIN_ID_HEX
+
+# Deploy to Dispatch
+forge create --rpc-url fuji-dispatch \
```
ashucoder9, line 23: same for this, pls add an example command and a demo contract maybe

content/academy/interchain-messaging/03-avalanche-starter-kit/05-foundry-quickstart.mdx
navillanueva: actually immediately after this section they have ICM basics where they deploy the contract, call a function, and read a state, so i updated this section to just explain what each thing does (forge create, cast send, cast call) but with no example (as i was looking through the easiest exmaple and it would just be a dup of what was to come). 

i think its clearer this way but let me know if you still see it valuable to include maybe a test command with some other function

martineckardt (approved):

PR #2463, opened 2025-05-15
add: modal window to pchain faucet

ashucoder9:
also resolves the error where rate limiting gets applied before proper validation

### DIFF
--- a/lib/rateLimit.ts
+++ b/lib/rateLimit.ts
@@ -17,15 +17,10 @@ const DEFAULT_OPTIONS: RateLimitOptions = {
   maxRequests: 1,
 };
 
-async function defaultIdentifier(req: NextRequest): Promise<string> {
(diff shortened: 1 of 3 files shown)


lib/rateLimit.ts
```diff
@@ -17,15 +17,11 @@ const DEFAULT_OPTIONS: RateLimitOptions = {
   maxRequests: 1,
 };
 
-async function defaultIdentifier(req: NextRequest): Promise<string> {
-  try {
+async function defaultIdentifier(): Promise<string> {
     const session = await import('@/lib/auth/authSession').then(mod => mod.getAuthSession());
```
martineckardt, line 21: Should be rather add this as an optional prob to the Toolbox app and hand it through the wallet to this component? Seems very hacks with dynamic imports

PR #2464, opened 2025-05-15
add: grant application forms for infrabuidl programs

ashucoder9:
includes both infraBUIDL and infraBUIDLAI


ashucoder9:
pls dnm yet, still waiting for a reply on whether to skip image uploads or not!

martineckardt (approved):

PR #2488, opened 2025-05-21
L1 Toolbox: Ledger fix / Ash Safe / WalletConnect support

owenwahlgren:
- align formatting of coreViem P-Chain methods to avoid repeated popups on failed txns (happens specifically on Ledger devices) ✅ 
- new extract P-Chain warp message methods in coreViem ✅ 
- decoupled steps in validator manager operations ✅ 
  - change weight ✅ 
  - remove validator ✅ 
  - add validator ✅ 
- Safe support with Ash ✅ 
- Safe support with Wallet Connect ✅ 
- some misc bug fixes and quality of life improvements

do not merge until 
- fully e2e tested with Safe Multisig + Core + a Ledger device ✅ 
- PoAManager is a real thing  ✅ 
- happy with UI/UX 🟨


toolbox/src/components/ConnectWallet/ConnectWallet.tsx
```diff
@@ -84,17 +83,47 @@ export const ConnectWallet = ({
         setIsClient(true)
     }, [])
 
-    useEffect(() => {
-        if (!walletEVMAddress || !walletChainId || !pChainAddress) return;
+    // Helper function to safely call Core wallet methods - only catch specific WalletConnect errors
```
nicolasarnedo99, line 86: Just out of curiosity - what was the requirement or issue to add support for WalletConnect? and where is this displayed, i didn't come across this functionality testing the demo

toolbox/src/components/MultisigOption.tsx
```diff
@@ -0,0 +1,468 @@
...
+
+  const getSupportedChain = async (chainId: string): Promise<string> => {
+    try {
+      const response = await fetch('/api/safe_chains');
+      
+      if (!response.ok) {
+        throw new Error(`HTTP ${response.status}: ${response.statusText}`);
+      }
+      
+      const data = await response.json();
```
nicolasarnedo99: is the data not already formatted to json in app/api/safe_chains/route.ts?

toolbox/src/components/MultisigOption.tsx
```diff
@@ -0,0 +1,468 @@
...
+      
+      onSuccess(safeTxHash);
+    } catch (err) {
+      onError(`Failed to propose transaction: ${(err as Error).message}`);
+    } finally {
+      setIsProposing(false);
+    }
+  };
+
+  const executeDirectTransaction = async () => {
```
nicolasarnedo99, line 294: Again double checking - this is for when the PoA owner address is not a multisig right?

toolbox/src/toolbox/hooks/useValidatorManagerDetails.ts
```diff
@@ -196,6 +224,106 @@ export function useValidatorManagerDetails({ subnetId }: UseValidatorManagerDeta
...
+        fetchContractOwner();
+    }, [validatorManagerAddress, publicClient]);
+
+    // Detect owner contract type when owner is a contract
+    useEffect(() => {
+        const detectOwnerType = async () => {
+            if (!isOwnerContract || !contractOwner || !publicClient) {
+                // If owner is not a contract but we have an owner, it's an EOA
+                if (!isOwnerContract && contractOwner) {
+                    setOwnerType('EOA');
```
nicolasarnedo99: is this not redundant with the first hook setting the Owner Type to an EOA if the owner is not a contract? I think that by isOwnerContract to true we dont not need to handle EOA cases anymore in this second hook

navillanueva (commented):
**Demo Coverage**
Setup L1 ✅
Setup Validator Manager ✅

Test add from single account ✅
Test remove from single account ✅
Test change weight from single account ✅

Deploy ASH multisig ✅
Deploy PoAManager and initialize it to ash multisig as owner ✅
Transfer ValidatorManager ownership to PoAManager ❌ Unable to calculate gas limit

Test add with multisig
Test remove with multisig
Test change weight with multisig

**Code Coverage**

Went into detail on most of the code except the following files that were mostly skimmed and will re-review once functionality is tested e2e:

- [ ] extractL1ValidatorWeightMessage
- [ ] extractRegisterL1ValidatorMessage
- [ ] refactored Add Validator Classes
- [ ] refactored Removed Validator Classes
- [ ] refactored Change Weight Classes

toolbox/src/components/ConnectWallet/ConnectWallet.tsx
owenwahlgren: walletConnect accounts will not have a pChainAddress that can be derived

toolbox/src/components/MultisigOption.tsx
owenwahlgren: this component will try to transact on the owner of ValidatorManager, which is a PoAManager.
The owner of PoAManager should be an Ash multisig.

If the connected account is WalletConnect, we check if PoAManager's owner is the WalletConnect account, in that case we can perform a normal tx and WalletConnect will forward it to Ash UI. 
If connected account is not owner of PoAManager, we will try to load the owner as an Ash multisig using the SafeSDK

toolbox/src/toolbox/hooks/useValidatorManagerDetails.ts
owenwahlgren: true, this `isOwnerContract` can be removed. We can prefer using the Owner Type.

toolbox/src/components/MultisigOption.tsx
owenwahlgren: good callout, I'll double check

toolbox/src/coreViem/methods/registerL1Validator.ts
```diff
@@ -69,13 +69,13 @@ export async function registerL1Validator(client: WalletClient<any, any, any, Co
     const unsignedRegisterValidatorTxHex = bytesToHex(unsignedRegisterValidatorTxBytes);
 
     // Submit the transaction to the P-Chain using Core Wallet
-    const txID = (await client.request({
+    const txID = await window.avalanche!.request({
```
containerman17, line 72: Why?

app/api/safe_chains/route.ts
```diff
@@ -0,0 +1,28 @@
+import { NextResponse } from 'next/server';
```
containerman17: This basically renders toolbox useless on it's own. Instead of fetching every time, let's bake it in and update in a post-install script.

toolbox/src/components/MultisigOption.tsx
owenwahlgren: when we make a fetch() call to it from the client side, we still need to call .json() on the response to parse the JSON body.
Server side: NextResponse.json(data) - formats the response as JSON
Client side: await response.json() - parses the JSON response body

containerman17:
Tried the flow on mainnet using this branch:

### ✅ Create Chain

All good as usual.

### ✅ Launch Node

I used the BlockScout component to launch a node — all good.

### ✅ Deploy VM on C-Chain

No issues.

### ✅ Init VM

Worked smoothly.

### ✅ Convert Subnet to L1

All good here too.

### 📝 Launching Another Node

All good, except for `AVAGO_HTTP_PORT` and `AVAGO_STAKING_PORT`. We should mention these even for the default ports, so users know what to change if they need non-default values. I needed to use non-default ports to run two nodes on the same machine.

### 📝 Adding a Validator (Regular Way)

Everything worked great! One small nit — the "Reset all steps" button feels like a primary call to action. It should probably be styled more like a secondary action.

### ✅ Change Weight (Regular Way)

All good. The P-Chain transaction failed the first time, but succeeded on retry.

### ✅ Remove Validator (Regular Way)

No issues.

### 📝 Query L1 Validator Set

Great tool! Just one thing — if you forget to fill out the `subnetId` field, it tries to load primary network validators and takes forever. Not sure what the best solution is for that.

### 📝 Deploy PoA Manager

All good. Would be helpful to include a link to the Ash wallet UI.

### ✅ Change Owner

Smooth and seamless.

### 📝 Add Validator via Ash Wallet

Overall, a great experience! Only feedback is that it wasn't obvious I could paste an EVM tx hash from Ash and continue the flow.

### ✅ Change Weight via Ash Wallet

Everything went great!

### ✅ Remove Validator via Ash Wallet

All good — had to retry the P-Chain transaction, but it was a minor issue.

containerman17 (changes requested):

app/api/safe_chains/route.ts
owenwahlgren: yeah it isn't ideal, but needed to avoid CORS error. The only side effect is that MultisigOption won't work on a toolbox standalone deployment.

toolbox/src/coreViem/methods/registerL1Validator.ts
owenwahlgren: to be consistent with styling in `convertSubnetToL1` method

martineckardt (dismissed):

app/api/safe_chains/route.ts
owenwahlgren: resolving for now

PR #2498, opened 2025-05-24
[Toolbox] Fix crash if WalletConnect is used and clean up EVM address lookup

martineckardt:


owenwahlgren:
this PR: https://github.com/ava-labs/builders-hub/pull/2488

also handles WalletConnect crashing

PR #2557, opened 2025-06-10
`SelectBlockchain` and refactored ChainID to BlockchainId

navillanueva:


toolbox/src/components/BlockchainDetailsDisplay.tsx
```diff
@@ -0,0 +1,181 @@
...
+    }
+
+    const copyToClipboard = (text: string | number | undefined | null) => {
+        if (!text) return
+        navigator.clipboard.writeText(text.toString())
+        setCopiedText(text.toString())
+        setTimeout(() => setCopiedText(null), 2000)
+    }
+
+    const isNonStandardVM = blockchain.vmId && blockchain.vmId !== STANDARD_EVM_VM_ID
```
martineckardt: Many AvaCloud chains have a different VM ID but are still standard Subnet-EVM chains

nicolasarnedo99: Maybe the phrasing is wrong, just wanted to identify those VM IDs that are not srEX.... but still use the standard subnet-evm

PR #2583, opened 2025-06-16
added custom VM ID extra steps

navillanueva:


toolbox/src/toolbox/Nodes/AvalanchegoDocker.tsx
```diff
@@ -61,8 +61,14 @@ export const nodeConfigBase64 = (chainId: string, debugEnabled: boolean, pruning
     return btoa(JSON.stringify(configMap))
 }
 
+const generateVMAliasesCommand = (vmId: string) => {
```
martineckardt: This can be removed because now using the flag

toolbox/src/toolbox/Nodes/AvalanchegoDocker.tsx
```diff
@@ -304,12 +341,52 @@ export default function AvalanchegoDocker() {
                                     <li><strong>9651</strong> (for the node-to-node communication)</li>
                                 </ul>
                             </Step>)}
+
+                            {/* Custom VM Setup Step */}
+                            {isCustomVM && (
```
martineckardt: Should not require an additional step now

martineckardt (changes requested):

martineckardt (approved):

PR #2609, opened 2025-06-30
add: Avalanche Ecosystem Stats page

ashucoder9:


navillanueva (commented):
UI looks great - im not so sure about showcasing inactive chains? they don't add to the total stats, might be useful to have a filter on the activity page and have them by default off, while the other are visible (high, medium, low)

app/(home)/stats/page.tsx
martineckardt, line 1: Please change to /stats/l1

martineckardt (changes requested):

martineckardt (approved):

PR #2626, opened 2025-07-07
Update rpc-providers.mdx to include Grove 🌿

fredteumer:
Adding Grove 🌿 to the list of RPC providers (both public and private endpoints).


owenwahlgren:
want to add yourself to [integrations](https://github.com/ava-labs/builders-hub/tree/master/content/integrations) as well?

PR #2632, opened 2025-07-09
update: add metrics from routescan

ashucoder9:
only ICM is not being fetched from routescan.


ashucoder9:
will add the cronjob for updating stats now

PR #2642, opened 2025-07-10
primary network avalanchego docker

owenwahlgren:


toolbox/src/toolbox/Nodes/AvalanchegoDocker.tsx
martineckardt, line 1: I am not a fan of mixing L1 Node Setup and Primary Network Node Setup for the following reasons:

**Intuitive Dev Ex:**
- A Primary Network node operator should not need to know what a L1/Subnet or the C-Chain is. They should not be confronted with the task to pick a BlockchainID. It is also not intuitive that you have to select the C-Chain, while the actual 
- There are plently of AvalancheGo flags that either apply to Primary Network nodes, L1 nodes or both. While I can set all flags on each node, we want to encourage not the two node types. A user should either set up a Primary Network or an L1 Node. This will reduce the options that apply to either node type.
- Having a dedicated component will enhance the experience with the AI Chatbot. Someone asking about Primary Network Nodes will get a more streamline tool just for that directly in the chat without unnecessary clutter.

**Recommend proper docker image:**
We should not promote using the Subnet-EVM image for Primary Network nodes. It has two version numbers (Subnet-EVM and AvalanchGo version). Primary Network only need to update when the AvalancheGo version changes, which is unecessary complexity. Also we the release of the Subnet-EVM image is a manual process, so there is no automation to build a new one once there is a new AvalancheGo version available. We may loose time in time-critical releases here. Furthermore, we don't want to expose Primary Network Validators to possible bugs in Subnet-EVM.

**Code Maintainability**
The Node Setup Tool is already too long (640 LOC) and we should split it in multiple sub components. To avoid code duplication we should abstract some components and reuse them across both tools. This is a good exercise anyway to reduce complexity. Some already are their own components like Docker Installation and the NodeReadinessValidator (we should rename this one). We should also separate these out:
- Ports to open step: Show list of ports that need to open based on RPC or Validator selection
- RPC Options: RPC Yes/No, if yes: Debug RPC?
- Reverse Proxy Setup
- Add to Wallet

PR #2734, opened 2025-08-13
[Toolbox] Console Redesign simple

martineckardt:


public/builderhub-console.png
nicolasarnedo99, line 1: image

looks to big compared to the dropdown of docs and grants for example

PR #2746, opened 2025-08-18
optimize data loading for validator stats page

ashucoder9:


app/api/c-chain-stats/route.ts
```diff
@@ -0,0 +1,421 @@
...
+
+async function getTimeSeriesData(metricType: string, pageSize: number = 365, fetchAllPages: boolean = false): Promise<TimeSeriesDataPoint[]> {
+  try {
+    let allResults: any[] = [];
+    let nextPageToken: string | null = null;
+    let pageCount = 0;
+    const maxPages = 50;
+    
+    do {
+      let url = `https://metrics.avax.network/v2/chains/${C_CHAIN_ID}/metrics/${metricType}?pageSize=${pageSize}&timeInterval=day`;
```
martineckardt: Why not use the AvalancheTypescriptSDK?

ashucoder9:
> * Use AvalancheTypscriptSDK
> * why do we need the API? We should just query from the metrics API from the frontend. Nothing should go through our backend

i'll add the new Avalanche SDK.

regarding the API stuff, using a backend proxy helps with protecting api keys, and preprocessing the data for display on the frontend. also we do multiple api calls via the proxy route, so accessing them requires lesser browser requests.

components/stats/ValidatorWorldMap.tsx
```diff
@@ -275,17 +275,23 @@ export function ValidatorWorldMap() {
...
             zoom={2}
             style={{ height: "100%", width: "100%" }}
-            scrollWheelZoom={true}
+            scrollWheelZoom={false}
+            zoomControl={false}
+            dragging={true}
+            touchZoom={false}
+            doubleClickZoom={false}
+            boxZoom={false}
+            keyboard={false}
```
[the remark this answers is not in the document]
ashucoder9: this is because we don't want map interactivity with keyboard, neither want to have zoom features or dragging (which creates overlapping map)

app/api/primary-network-stats/route.ts
```diff
@@ -1,381 +1,257 @@
...
-async function getValidatorWeight30Days() {
+async function getTimeSeriesData(metricType: string, timeRange: string, pageSize: number = 365, fetchAllPages: boolean = false): Promise<TimeSeriesDataPoint[]> {
   try {
-    const url = `https://metrics.avax.network/v2/networks/mainnet/metrics/validatorWeight?pageSize=30&subnetId=${PRIMARY_NETWORK_SUBNET_ID}`;
-    const response = await fetch(url, {
-      headers: { 'Accept': 'application/json' },
+    const { startTimestamp, endTimestamp } = getTimestampsFromTimeRange(timeRange);
+    let allResults: any[] = [];
+    let pageCount = 0;
+    const maxPages = 50;
```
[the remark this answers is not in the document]
ashucoder9: fixed.

0xstt (approved):

app/(home)/stats/primary-network/c-chain/page.tsx
martineckardt, line 1: Do we need this extra page or can we just use a regular L1 page?

ashucoder9: it's better to have c-chain under primary-network. if needed, we can use a regular L1 page too.

ashucoder9: added a shared component for c-chain and L1's stats

PR #2789, opened 2025-08-28
added ERC-7702 & Agentkit

Adi-21:
I have updated the following things that I have mentioned in the PR and they are,

0xGasless.mdx file have been updated with ERC-7702 & Agentkit that we are now supporting.

Also, I have wrapped layout.tsx file code with (Sessionprovider ) since i was facing issue when I was accessing Account Abstraction page and got useSession() error.


app/integrations/layout.tsx
```diff
@@ -1,15 +1,22 @@
+"use client";
+
 import { HomeLayout } from 'fumadocs-ui/layouts/home';
 import type { ReactNode } from 'react';
 import { Footer } from '@/components/navigation/footer';
 import { baseOptions } from '@/app/layout.config';
+import { SessionProvider } from 'next-auth/react';
```
ashucoder9: why do you need session provider in integrations page?

Adi-21: when I was going to account abstraction/ 0xgasless page then I was getting this error 
Screenshot 2025-08-28 at 2 47 10 PM

nicolasarnedo99:
Please stash the `bun.lock` changes as well, we use yarn

owenwahlgren (approved):

PR #2796, opened 2025-08-31
[Console] Extract managed testnet nodes logic into service and hook

martineckardt:


owenwahlgren:
https://github.com/ava-labs/builders-hub/pull/2988

PR #2839, opened 2025-09-08
`Codebase Entrepreneur Academy`: fixing certificates + improving flashcard to quiz user flow

navillanueva:


public/certificates/Codebase_EntrepreneurAcademy_Certificate.pdf
martineckardt, line 1: Can you store the certifcates in the Vercel blob storage instead of here? Please also move the other certificates

PR #2844, opened 2025-09-09
feat: refactor BuilderHub AI chat, migrate toolbox->console

owenwahlgren:


app/api/chat/console-tools.ts
```diff
@@ -0,0 +1,97 @@
+export const CONSOLE_TOOLS = {
```
martineckardt: Can we reuse the tool list from the console-sidebar here?

app/api/chat/prompt.ts
```diff
@@ -0,0 +1,51 @@
...
+    Answer policy (verbosity by default):
+    - Default to detailed answers (5–12 sentences). Use short sections, bullets, and code when helpful.
+    - Include 2–4 short quotes with source links when giving technical guidance.
+    - Provide a compact step-by-step or checklist when users need to execute tasks.
+
+    Safe links (only use these or URLs explicitly present in context "Source URL:"):
+    - Console: https://build.avax.network/console
+    - Documentation: https://build.avax.network/docs
+    - Academy: https://build.avax.network/academy
+    - Guides: https://build.avax.network/guides
```
martineckardt: No more guides, use blog

components/ai/console-renderer.tsx
```diff
@@ -0,0 +1,165 @@
+'use client';
+import { ChevronRight, Loader2, X } from 'lucide-react';
+import dynamic from 'next/dynamic';
+import { useEffect, useState } from 'react';
+import React from 'react';
+
+// Console component mapping - maps console paths to their actual components
+const consoleComponentMap: Record<string, () => Promise<{ default: React.ComponentType<any> }>> = {
```
martineckardt: Extract in own file and add to console directory, so we don't miss adding them here when we add new tools

app/api/chat/console-tools.ts
owenwahlgren: no because it is on frontend. We can make a separate constant file which both sidebar and backend can pull from.

PR #2956, opened 2025-09-17
update: search provider to algolia

ashucoder9:


components/search.tsx
```diff
@@ -1,33 +1,46 @@
...
+import { useState } from "react";
+import algoliasearch from "algoliasearch/lite";
+import { useDocsSearch } from "fumadocs-core/search/client";
 
-const trieveClient = new TrieveSDK({
-    apiKey: 'tr-DRwBhkLLDJGjRfVk3T9PgIPgVrcE2oKs',
-    datasetId: '4948db5c-a7e7-423f-bf6a-0ac1ba7a2b76',
-});
+const appId = "0T4ZBDJ3AF";
+const apiKey = "9b74c8a3bba6e59a00209193be3eb63a";
```
nicolasarnedo99, line 8: should we move this to .env?

ashucoder9: no, we don't have to. it's public.

navillanueva (approved):

PR #2985, opened 2025-09-19
add: notifications to faucet claims

ashucoder9:


components/toolbox/components/ConnectWallet/EVMFaucetButton.tsx
```diff
@@ -46,6 +48,14 @@ export const EVMFaucetButton = ({
...
 
+    notify(
+      {
+        type: "local",
+        name: `${chainConfig.coinName} Faucet Claim`,
+      },
+      faucetRequest
+    );
+
     consoleToast.promise(faucetRequest, {
```
owenwahlgren: notify() will handle promise completely, no need for consoleToast anymore

components/toolbox/components/ConnectWallet/PChainFaucetButton.tsx
```diff
@@ -50,7 +52,17 @@ export const PChainFaucetButton = ({ className, buttonProps, children }: PChainF
...
+
+    notify(
+      {
+        type: "local",
+        name: "P-Chain AVAX Faucet Claim",
+      },
+      faucetPromise
+    );
+
+    consoleToast.promise(faucetPromise, {
```
owenwahlgren: same here
notify() will handle promise completely, no need for consoleToast

components/toolbox/components/ConnectWallet/EVMFaucetButton.tsx
ashucoder9: right now, the notification toasts also have custom functions and errors like: showing Rate Limit messages, and Logic button as well as explorer buttons. so either we standardize this, or we keep it as it is rn for faucet scope.

components/toolbox/components/ConnectWallet/PChainFaucetButton.tsx
ashucoder9: same concern as above.

owenwahlgren (approved):

PR #3014, opened 2025-09-24
Enhance log fetching range and UI feedback

owenwahlgren:
Updated the block range for fetching logs from the last 2,000 to the last 100,000 blocks. Improved the user interface for fetching events with progress indication.


Hardikgosai1 (commented):
Supeeerrrseupeeeerr

Hardikgosai1 (commented):
Superb

PR #3119, opened 2025-10-17
Continue button form submmit hackathons fixed

alejandro99so:


navillanueva (dismissed):

Andyvargtz:
Added some comments in the Vercel about the mandatory fields not being highlighted in order to continue, so continue button doesn't perform action but no user feedback is included.

alejandro99so:
comments from vercel applied

PR #3129, opened 2025-10-20
sidebar active item and color contrast on sidebar

federiconardelli7:
fix the active item on sidebar for multisteps components in console
modified the color of sidebar active to enhance contrast on both themes


federiconardelli7:
image
image

PR #3174, opened 2025-10-24
`Academy`: added embedded view + nits

navillanueva:


components/toolbox/hooks/useAccountRequirements.ts
```diff
@@ -54,8 +54,15 @@ export function useAccountRequirements(configKey: AccountRequirementsConfigKey |
...
-    const status = session?.status || 'loading';
+    // Safely get session - if SessionProvider is not available, default to unauthenticated
+    let session;
+    try {
+        session = useSession();
+    } catch (error) {
+        // SessionProvider not available in this context
+        session = { status: 'unauthenticated' as const, data: null };
+    }
+    const status = session?.status || 'unauthenticated';
```
nicolasarnedo99: @owenwahlgren added this because i was getting issues when testing on lcoalhost but i think it might not be necessary correct me on it

owenwahlgren: I think this may be unsecure cc @0xstt

0xstt: You experienced these issues in the Academy content, right? It’s probably because there’s no `SessionProvider` wrapping our tools. Why not add `SessionProvider` to the `ToolboxMdxWrapper`?

PR #3207, opened 2025-10-28
[FIX]:Enhance profile access control in API

SebasianDev:
added checks to ensure users can only access and update their own profiles, with improved error handling for unauthorized requests.


Andyvargtz:
Changes migrated to #3209

PR #3264, opened 2025-11-04
Recovery solidity course

alejandro99so:


Andyvargtz (changes requested):
Please move solidity course below Avalanche fundamentals in the tree.

Andyvargtz (approved):

PR #3271, opened 2025-11-05
comment out point system for badges requirements in Staging_badges + comment out hackathon logic + change to achievements board in ui

katherineavalabs:
Commenting with mark COMMENTED OUT: Points feature disabled for future reference


nicolasarnedo99:
@katherineavalabs can we just directly delete it instead of commenting it out?

PR #3337, opened 2025-11-14
improve header stat page

federiconardelli7:
- add snowscan for c-chain as block explorer
- add the header for the stat page.

final version:
image
image

first commit version that includes the 7 cards (as possible alternative if you like more) :
image

### DIFF
--- a/app/(home)/stats/overview/page.tsx
+++ b/app/(home)/stats/overview/page.tsx
@@ -408,10 +408,7 @@ export default function AvalancheMetrics() {
                     "_blank"
                   )
                 }
-                className="flex-shrink-0 text-white"
-                style={{
-                  background: 'linear-gradient(135deg, #E84142, #3752AC)'
-                }}
+                className="flex-shrink-0 border border-white bg-black text-white hover:bg-neutral-800 dark:bg-transparent dark:hover:border-neutral-400 transition-colors"
               >
                 Submit Your L1
                 <ArrowUpRight className="ml-1.5 h-4 w-4" />


ashucoder9:
the design which includes the 7 cards looks better @federiconardelli7

federiconardelli7:
> the design which includes the 7 cards looks better @federiconardelli7

Sounds good, I'll revert the commit. Do you like the button gradient color or should I make like the other version?

federiconardelli7:
Reverted back to the header+hero with 7 cards included.

app/(home)/stats/overview/page.tsx
```diff
@@ -363,139 +363,162 @@ export default function AvalancheMetrics() {
...
+                size="sm"
+                onClick={() =>
+                  window.open(
+                    "https://github.com/ava-labs/builders-hub/blob/master/constants/l1-chains.json",
+                    "_blank"
+                  )
+                }
+                className="flex-shrink-0 text-white"
+                style={{
+                  background: 'linear-gradient(135deg, #E84142, #3752AC)'
```
nicolasarnedo99: Imo a gradient button on an already gradient banner looks a bit too much, can we go from something simpler like white button

federiconardelli7: Yes, I would like to make like the other version, so black or white (based on the theme) with an on-hover event

martineckardt (dismissed):

meaghanfitzgerald:
I can't catch it in a screenshot, but the loading image for the page seems off to me. It doesn't reflect the information being loaded, and flashes when the info finally loads. This should be changed to reflect the actual page layout

martineckardt (approved):

federiconardelli7:
image
@meaghanfitzgerald just catch it (slow computers are useful for something ahah)

I'll work on that

PR #3343, opened 2025-11-17
fix loading screen text on C-chain and L1s

federiconardelli7:
Fix the text while loading the C-chain page and L1 chains 

image

image

### DIFF
--- a/components/ui/granite-banner.tsx
+++ b/components/ui/granite-banner.tsx
@@ -1,6 +1,6 @@
 "use client";
 import { Banner } from "fumadocs-ui/components/banner";
-import { useEffect, useState } from "react";
+import { useEffect, useState, useRef } from "react";
 import Link from "next/link";
 
 interface TimeUnit {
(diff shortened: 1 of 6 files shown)


meaghanfitzgerald:
this is the loading screen im seeing 
image

meaghanfitzgerald:
c-chain page loading still looks odd, has a bullet point in front of the title and has a different aspect ratio than the loaded state it seems

federiconardelli7:
image

Now it should be fixed for the C-chain and L1 pages. 

Pushing when it goes up again.

nicolasarnedo99:
@federiconardelli7 can this be turned into a component/wrapper that we apply to all three loading stats? code looks repeated for the c-chain and the l1 page

federiconardelli7:
> @federiconardelli7 can this be turned into a component/wrapper that we apply to all three loading stats? code looks repeated for the c-chain and the l1 page

Yes and no, because it's the actual skeleton of what the page loaded looks like. So the overview page is different from the validator page and the chain page.
The C-chain and l1 page is the same (ChainMetricsPage.tsx), it automatically adjusts if there is or there is not the logo, displays the fixed texts, and blurs where the text needs to be displayed

PR #3345, opened 2025-11-18
create stats/playground page

0xstt:


prisma/schema.prisma
```diff
@@ -220,4 +222,34 @@ model ConsoleLog {
 
   @@index([user_id])
   @@index([created_at])
+}
+
+model Playground {
```
martineckardt: Should we be more specific? Like `StatsPlayground`

PR #3356, opened 2025-11-19
add: icm stats page

ashucoder9:


containerman17 (approved):

app/(home)/stats/interchain-activity/page.tsx
martineckardt, line 1: should be `/stats/interchain-messaging`

PR #3375, opened 2025-11-22
Fix syntax error in mapping assignment

naiih001:
The assignment was a bit buggy


naiih001:
Thanks for the review

PR #3401, opened 2025-11-25
Fix subnetID parameter placement

muhammetselimfe:
The platform.getCurrentValidators RPC playground was incorrectly placing
  subnetID outside the params object. When users entered a subnet ID in the
  UI, it appeared at the root level instead of inside params, causing the
  API to ignore the subnet filter and return primary network validators.

  **Root Cause:**
  The bug was in the custom React component (components/api/rpc-api-page.client.tsx),
  The BodyFieldWithExpandedParams component was
  flattening nested field values instead of maintaining the proper nested structure.

  **Fix:**
  Updated the component to correctly handle nested field paths:
  - Fixed the value getter to traverse nested paths (e.g., params.subnetID)
  - Fixed the onChange handler to create proper nested object structures
  - Ensures all params properties are correctly placed inside the params object

  This fixes the issue for platform.getCurrentValidators and any other RPC
  methods using nested params.

### DIFF
--- a/components/api/rpc-api-page.client.tsx
+++ b/components/api/rpc-api-page.client.tsx
@@ -75,13 +75,41 @@ export function BodyFieldWithExpandedParams({
                 id={`body.${fieldPath}`}
                 placeholder={propSchema.default || "Enter value"}
                 type={propSchema.type === 'integer' || propSchema.type === 'number' ? 'number' : 'text'}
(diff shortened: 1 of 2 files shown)


owenwahlgren (approved):

owenwahlgren:
thank you!

owenwahlgren:
needs verified commits, closing for https://github.com/ava-labs/builders-hub/pull/3404

PR #3419, opened 2025-11-27
update: ui improvement on the API pages

ashucoder9:


components/api/api-open-body.client.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+// automatically open the Body section on OpenAPI pages
+export function OpenBodySection() {
+  useEffect(() => {
+    const openBodySection = () => {
+      const buttons = Array.from(document.querySelectorAll("button"));
+
+      buttons.forEach((button) => {
+        if (
+          button.textContent?.trim().startsWith("Body") &&
```
containerman17: A bit too hacky, can  you just add a class/id to the button instead of relying on text?

app/docs/styles.css
```diff
@@ -445,3 +445,7 @@ aside[aria-label*="On this page" i] > div:first-child,
   }
 }
 
+/* Add extra horizontal padding to the first button in response tablist */
+div [role="tablist"] button:first-child {
+  padding-left: 0.5rem !important;
```
containerman17, line 488: Just make sure you really need `!important` - it causes pain sometimes

components/api/api-open-body.client.tsx
ashucoder9: Since the panel buttons are rendered by fumadocs-openapi (library), we can't directly add a class/id. Instead, we are now marking the buttons with a data attribute (data-playground-panel-trigger) when first found and reuse the data attribute for subsequent selections. changes in the new commit.

PR #3426, opened 2025-11-27
Explorer pages

0xstt:


meaghanfitzgerald:
can we add a shortcut to the navbar dropdown 
image

meaghanfitzgerald:
the explorer and other "stats" pages have this navbar always blocking the "load more" or final row of info on the display page 

https://github.com/user-attachments/assets/d56e2e08-3367-4dd1-ab0f-88df85ce08e1

0xstt:
> the explorer and other "stats" pages have this navbar always blocking the "load more" or final row of info on the display page
> 
>  Screen.Recording.2025-12-01.at.10.43.50.AM.mov

Hey thanks @meaghanfitzgerald, I’m creating an issue for that since these are already existing problems.

app/(home)/stats/network-metrics/page.tsx
martineckardt, line 1: I think we should call the url something more descriptive. For example `/stats/network-metrics

martineckardt (dismissed):

PR #3434, opened 2025-11-28
Bugfix/hackathons pagination

simon0820s:
Fix bug: 3387

Apply status filters directly at DB level (ENDED / ONGOING / UPCOMING) using Prisma where conditions.
Integrate searchFilters into the main query by pushing { OR: searchFilters } into the shared conditions array.
Remove redundant in-memory status filtering over hackathons after fetching from DB.
Drop the extra findMany + getHackathonLite roundtrip previously used only to recompute totalHackathons by status.
Simplify the response by deriving total from hackathonsLite.length, aligned with the already filtered in-memory collection.


server/services/hackathons.ts
```diff
@@ -306,7 +265,7 @@ export async function getFilteredHackathons(options: GetHackathonsOptions) {
           ),
         } as HackathonHeader)
     ),
-    total: totalHackathons,
+    total: hackathonsLite.length,
```
Another-DevX: There might be some bugs here due to faulty pagination.

mathecodemoon:
Fixed !!

Andyvargtz:
@simon0820s please verify your commits so we can merge this one

PR #3448, opened 2025-12-01
add: more page for L1 validators

ashucoder9:


app/(home)/stats/validators/page.tsx
```diff
@@ -105,11 +111,7 @@ export default function ValidatorStatsPage() {
     const extractNumbers = (v: string) => {
       const match = v.match(/(\d+)\.(\d+)\.(\d+)/);
       if (!match) return [0, 0, 0];
-      return [
```
containerman17, line 108: Why wouldn't you just treat it as string? Why parsing?

PR #3457, opened 2025-12-02
feat/orama search

owenwahlgren:
https://fumadocs.dev/docs/ui/search/orama

should reduce amount of self inflicted 404s from Agolia.

image
image
image


ashucoder9:
i think updating the search index (just did) with recent docs changes should fix the 404 issue. algolia is far better for search than orama, but less customizable. also do check the new search suggestions, should result in lesser 404s

PR #3460, opened 2025-12-02
fixing links

navillanueva:


components/landing/development.tsx
```diff
@@ -13,7 +13,7 @@ const development = [
     description:
       "Find references for all Avalanche SDKs, APIs, and tools.",
     icon: Book,
-    href: "/docs/dapps",
+    href: "/docs",
```
owenwahlgren, line 16: /docs doesn't exist technically (no index file), its /docs/primary-network as landing

next.config.mjs
```diff
@@ -86,6 +86,56 @@ const config = {
...
         permanent: true,
       },
+      {
+        source: '/docs/nodes/validate/validate-vs-delegate',
+        destination: '/docs/primary-network/validate/validate-vs-delegate',
+        permanent: true,
+      },
+      {
+        source: '/docs/dapps/end-to-end/fuji-workflow',
+        destination: '/blog/what-is-fuji-testnet',
```
owenwahlgren: this prolly isn't best alternative link. can we link to academy?

owenwahlgren (approved):

next.config.mjs
owenwahlgren: like here /academy/blockchain/solidity-foundry/04-hello-world-part-1/01-intro

PR #3463, opened 2025-12-02
patch to redirect paths

alejandro99so:
patch to redirect paths not valid from academy, docs and console


owenwahlgren:
I'd push back on this approach - silently redirecting to parent paths masks broken links rather than surfacing them. We lose visibility into issues like outdated links, deleted content, or typos in URLs since they never show up as 404s in our logs.

Instead, we should keep the 404 page and add a "Did you mean?" suggestion button that links to the nearest parent path. That way users still get a helpful nudge, but we maintain visibility into what's broken.

PR #3532, opened 2025-12-07
docs: add Solidity compiler evmVersion requirement for Avalanche

owenwahlgren:
## Summary
- Documents the need to set Solidity compiler `evmVersion` to `cancun` when deploying to Avalanche
- Avalanche supports Cancun but not Pectra (which is the default since Solidity v0.8.30)
- This is a temporary requirement that will be removed when Avalanche adds Pectra support after SAE implementation

## Changes
- **deploy-smart-contract.mdx**: Added warning callout with Remix-specific instructions
- **hardhat.mdx**: Added warning callout and updated config example with `evmVersion: "cancun"`
- **foundry-quickstart.mdx**: Added `evm_version = "cancun"` to foundry.toml example

## Context
From the [subnet-evm README](https://github.com/ava-labs/subnet-evm#compatibility):
> Subnet-EVM and Avalanche C-Chain currently implement the Ethereum Cancun fork and do not yet support newer hardforks (such as Pectra). Since Solidity v0.8.30 switched its default target EVM version to Pectra, contracts compiled with default settings may emit bytecode using instructions/features that Avalanche does not support.

## Test plan
- [x] Verified all three pages render correctly (HTTP 200)
- [x] Dev server runs without errors


meaghanfitzgerald (approved):
To market we go

PR #3545, opened 2025-12-08
update the hook to use internal explorer for the current supported chain

federiconardelli7:


hooks/useEVMNotifications.ts
```diff
@@ -5,16 +5,37 @@ import { Chain, createPublicClient, http } from 'viem';
...
-    } else if (viemChain.id === 43114 || viemChain.id === 43113) {
-        return `https://${viemChain.id === 43113 ? "subnets-test" : "subnets"}.avax.network/c-chain/tx/${txHash}`;
-    } else {
-        const rpcUrl = viemChain.rpcUrls.default.http[0];
-        return `https://devnet.routescan.io/tx/${txHash}?rpc=${rpcUrl}`;
     }
+    
+    // Last resort: use routescan with RPC URL
+    const rpcUrl = viemChain.rpcUrls.default.http[0];
+    return `https://devnet.routescan.io/tx/${txHash}?rpc=${rpcUrl}`;
```
owenwahlgren, line 38: we can use builderhub explorer if chain is added to console!

federiconardelli7: Yes, now it should work like that. First it check for the static L1 and then the added from the console. If they can't find it try just as last resort with the routescan explorer 
image

If you want I can delete the last resort or simply force it with the builder-hub explorer

PR #3566, opened 2025-12-11
fix(homepage): resolve canvas blur and horizontal scroll issues

owenwahlgren:
- Add devicePixelRatio (DPR) support for sharp rendering on Retina/HiDPI displays
- Scale canvas buffer to device resolution while keeping CSS size at logical pixels
- Use ctx.resetTransform() + ctx.scale(dpr) pattern for correct coordinate mapping
- Fix horizontal scroll by changing nebula gradient from 200vw/200vh to 200%
- Add overflow-hidden to container to ensure gradient stays contained

Fixes regression introduced in PR #3535

🔺 Generated by DevRel swarm

### DIFF
--- a/components/landing/globe.tsx
+++ b/components/landing/globe.tsx
@@ -252,7 +252,7 @@ export const Sponsors = () => {
 	}
 
 	return (
-		<div className="relative w-full h-[700px] flex items-center justify-center">
+		<div className="relative w-full h-[760px] flex items-center justify-center">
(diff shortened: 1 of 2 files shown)


owenwahlgren:
@copilot the buttons to view stats and explorer are getting cut off the homepage globe animation.

owenwahlgren:
@copilot I can still scroll side to side

PR #3615, opened 2025-12-19
Fix validator stats + Remove validators page

0xstt:


martineckardt:
We need to update the "validators" link in the navigation. It still points to the hidden page

PR #3645, opened 2026-01-06
removed blockjoy from integrations. No longer in operation

federiconardelli7:


owenwahlgren:
can you remove the image too please?

PR #3691, opened 2026-01-14
fix(console): set allow-unfinalized-queries to false by default

owenwahlgren:
## Summary
- Changed the default value for `allow-unfinalized-queries` from `true` to `false` for both **Public RPC** and **Validator+RPC** node types in the L1 Node Setup wizard

## Why
Setting `allow-unfinalized-queries: true` lets RPC nodes return data from blocks that haven't been finalized yet. 

Setting the default to `false` ensures users receive only finalized/confirmed block data, which is safer for production applications. Users who specifically need real-time unfinalized queries can still enable this option manually via the checkbox in Advanced Settings.

## Test plan
- [ ] Navigate to Console → L1 Node Setup
- [ ] Select "Public RPC Node" type → verify `allow-unfinalized-queries` is unchecked by default
- [ ] Select "Validator + Public RPC" type → verify `allow-unfinalized-queries` is unchecked by default
- [ ] Select "Validator Node" type → verify behavior unchanged (was already `false`)
- [ ] Toggle the checkbox in Advanced Settings → verify it still works correctly

### DIFF
--- a/components/toolbox/console/layer-1/AvalancheGoDockerL1.tsx
+++ b/components/toolbox/console/layer-1/AvalancheGoDockerL1.tsx
@@ -157,7 +157,7 @@ function AvalanchegoDockerInner() {
             // RPC node defaults - optimized for query performance
             setPruningEnabled(false); // RPC nodes typically need full history
             setLogLevel("info");
-            setAllowUnfinalizedQueries(true); // Enable real-time queries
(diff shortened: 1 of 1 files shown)


owenwahlgren:
Closing - branch was created from wrong base, includes unrelated changes

PR #3743, opened 2026-01-25
fix(console): show requirements checklist on data-api-keys page

federiconardelli7:
- Fix UX inconsistency where data-api-keys was the only console tool redirecting to login instead of showing the requirements checklist                                     
  - Create `/api/glacier-jwt` endpoint for client-side JWT fetching                                                                                                           
  - Wrap TokenManagement with `withConsoleToolMetadata` HOC using `AccountRequirementsConfigKey.UserLoggedIn` 
  
  ## Test plan                                                                                                                                                                
  - [x] Navigate to `/console/utilities/data-api-keys` while logged out                                                                                                       
  - [x] Verify "To use this tool you need:" modal appears with "Builder Hub Account" requirement                                                                              
  - [x] Click "Log In" and verify login modal opens                                                                                                                           
  - [x] After logging in, verify tool loads and API keys list works                                                                                                           
  - [x] Test create/delete API key functionality

### DIFF
--- a/components/build-games/ReferralButton.tsx
+++ b/components/build-games/ReferralButton.tsx
@@ -1,19 +1,55 @@
 "use client";
 
 import { useState, useEffect } from "react";
+import { useSession, getSession } from "next-auth/react";
 import ReferralModal from "./ReferralModal";
 import { captureReferrerFromUrl } from "@/lib/referral";
+import { useLoginModalTrigger } from "@/hooks/useLoginModal";
 
 export default function ReferralButton() {
   const [isModalOpen, setIsModalOpen] = useState(false);
+  const { data: session, status } = useSession();
+  const { openLoginModal } = useLoginModalTrigger();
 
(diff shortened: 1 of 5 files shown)


app/api/glacier-jwt/route.ts
```diff
@@ -0,0 +1,32 @@
...
+  const session = await getAuthSession();
+
+  if (!session) {
+    return NextResponse.json({ error: "Unauthorized" }, { status: 401 });
+  }
+
+  const DATA_API_ENDPOINT =
+    process.env.VERCEL_ENV === "production"
+      ? "https://data-api.avax.network/v1"
+      : "https://data-api-dev.avax.network/v1";
```
owenwahlgren: I think we can just use the prod data api endpoint for both :)

federiconardelli7: Updated with only prod data api for both

PR #3794, opened 2026-02-06
feat (console): enhanced performance monitor with table view

navillanueva:


owenwahlgren:
@copilot can we refactor out the RPC methods admin, platform, debug, eth etc to their own separate files and then use them in `components/toolbox/console/layer-1/PerformanceMonitor/RPCMethodsChecker.tsx`

PR #3819, opened 2026-02-10
fix(SBP-001,SBP-002): prevent hackathon admin API abuse by enforcing session-derived authz, sanitize markdown to prevent XSS

Another-DevX:


app/api/hackathons/route.ts
alejandro99so, line 16: Why is the ID not gotten from the user session that is calling the API as "POST", but without role locking the process?

components/hackathons/hackathon/sections/About.tsx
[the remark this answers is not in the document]
alejandro99so: The same in "components/hackathons/hackathon/sections/Submission.tsx#L113" <MDXRemote source={hackathon.content.judging_guidelines} />

app/api/hackathons/route.ts
Another-DevX: Fixed!

PR #3832, opened 2026-02-13
feat(academy): improve quizzes

navillanueva:


owenwahlgren:
12k+ lines in `quizData.json` maybe it is time we separate them all out?

PR #3912, opened 2026-03-02
fix: pass isTestnet flag when adding networks to Core wallet

owenwahlgren:
## Summary
- Testnet L1s were being added as mainnet networks in Core wallet
- **AddChainModal**: used viem's `addChain()` which only forwards standard EIP-3085 fields (`chainId`, `chainName`, `nativeCurrency`, `rpcUrls`, `blockExplorerUrls`), silently dropping Core wallet's proprietary `isTestnet` field. Replaced with a direct `wallet_addEthereumChain` request that includes `isTestnet`.
- **useNetworkActions**: the raw EIP-1193 fallback for `wallet_addEthereumChain` (triggered when wallet doesn't recognize the chain) omitted `isTestnet` entirely.

## Test plan
- [ ] Add a Fuji testnet L1 via the Add Chain modal → verify it appears under testnet in Core wallet
- [ ] Switch to an unknown testnet L1 via the console header network dropdown → verify Core adds it as testnet


meaghanfitzgerald (approved):
❌ commitlint ❌

PR #3952, opened 2026-03-11
fix(console): fix cross-chain transfer timeout and auto-import UX

owenwahlgren:
## Summary
- **False timeout bug**: P-Chain tx confirmation was timing out after 3s (10 polls × 300ms). Increased to 60s (30 polls × 2s) to match real-world P-Chain latency
- **Auto-import**: After export succeeds and UTXOs arrive, the import transaction now fires automatically — no extra button click needed. Core Wallet still prompts for signing
- **Step collapse timing**: Step 1 now collapses when step 2 is actionable (UTXOs arrived), not on an arbitrary 2s timer
- **Polling cleanup**: Replaced exponential backoff (1s, 2s, 3s...) with consistent 2s intervals, 15 attempts
- **Direction swap**: Auto-swap now only fires on initial page load, preventing mid-transfer confusion
- **Error recovery**: Import errors show inline retry button

## Test plan
- [x] C→P transfer: verify export → auto-import flow (wallet prompts twice, no extra button)
- [x] P→C transfer: same flow in reverse
- [x] Pre-existing UTXOs: verify manual import button shows with "pending from previous transfer" label
- [x] Slow P-Chain: verify no false timeout errors (wait up to 60s)
- [x] Import failure: verify retry button appears inline
- [x] Page reload mid-transfer: verify direction auto-detects correctly on load only

### DIFF
--- a/components/toolbox/console/primary-network/CrossChainTransfer.tsx
+++ b/components/toolbox/console/primary-network/CrossChainTransfer.tsx
@@ -236,7 +236,7 @@ function CrossChainTransfer({
                     fromAddress: walletEVMAddress as `0x${string}`
                 });
                 const txnResponse = await coreWalletClient.sendXPTransaction(txnRequest);
-                await coreWalletClient.waitForTxn(txnResponse);
+                await coreWalletClient.waitForTxn({ ...txnResponse, sleepTime: 2000, maxRetries: 30 });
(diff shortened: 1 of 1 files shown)


federiconardelli7:
Tested C -> P bridge and viceversa. Refreshed the page meanwhile and still fetched and worked correctly.

meaghanfitzgerald (approved):

PR #3954, opened 2026-03-12
add: Oauth for Team1 website

ashucoder9:
Builder Hub Oauth integration for Team1 Member website. Team1 Members can be fetched directly via their builder hub accounts, and can be later used for additional stuff like badges etc.


owenwahlgren (changes requested):
Hey @ashucoder9 — a couple of things before we can move forward:

1. **Please add a PR description** with context on what this OAuth integration does, why it's needed, and how it works. The PR body is currently empty.

2. **Have the database migrations been run?** If this OAuth flow requires any DB schema changes (new tables, columns, etc.), please confirm whether migrations have been applied to staging/prod or if they still need to be run.

Thanks!

PR #4005, opened 2026-03-23
feat: validator alerts (ditto + rest implementation)

ashucoder9:
changes from https://github.com/ava-labs/builders-hub/pull/4001 + completes the implementation of Validator Notify feature

1. db migration live on preview
2. env vars and keys set along with cron job for routine checkup (right now at 15 mins)

Next: test on preview + perform prod db migration

### DIFF
--- a/server/services/validator-alert-check.ts
+++ b/server/services/validator-alert-check.ts
@@ -314,19 +314,19 @@ export async function checkSingleAlert(
       );
       if (didSend) sent++;
     } else if (latestRelease.type === 'mandatory') {
+      // No parseable deadline — don't escalate to mandatory tier.
+      // Send as optional with a note that it may be mandatory.
       const didSend = await trySend(
         alert.id,
         alert.email,
-        'version_mandatory',
-        versionMandatoryTemplate({
+        'version_optional',
+        versionOptionalTemplate({
           alertId: alert.id,
           nodeId: alert.node_id,
           label: alert.label,
           currentVersion: validator.version,
-          requiredVersion: latestRelease.tag,
-          deadline: null,
-          acps: latestRelease.acps,
-          urgency: 'notice',
+          latestVersion: latestRelease.tag,
+          maybeMandatory: true,
         }),
         errors,
         alert.node_id
--- a/server/templates/validator-alerts.ts
+++ b/server/templates/validator-alerts.ts
@@ -177,24 +177,36 @@ export function versionOptionalTemplate(params: {
   label: string | null;
   currentVersion: string;
   latestVersion: string;
+  maybeMandatory?: boolean;
 }): { subject: string; html: string; text: string } {
   const name = params.label ? `${params.label} (${params.nodeId})` : params.nodeId;
-  const subject = `AvalancheGo ${params.latestVersion.replace('avalanchego/', 'v')} Available (Optional)`;
-  const text = `Your validator ${name} is running ${params.currentVersion}. A new optional release ${params.latestVersion} is available.`;
+  const mandatoryNote = params.maybeMandatory
+    ? ' — this release may be mandatory, check release notes for details'
+    : '';
+  const subject = params.maybeMandatory
+    ? `AvalancheGo ${params.latestVersion.replace('avalanchego/', 'v')} Available (Possibly Mandatory)`
+    : `AvalancheGo ${params.latestVersion.replace('avalanchego/', 'v')} Available (Optional)`;
+  const text = `Your validator ${name} is running ${params.currentVersion}. A new release ${params.latestVersion} is available${mandatoryNote}.`;
+  const heading = params.maybeMandatory
+    ? 'A new release is available — this release may be mandatory, check release notes for deadline'
+    : 'A new optional release is available (recommended but not required)';
+  const footnote = params.maybeMandatory
+    ? 'This release may be mandatory but no deadline could be determined. Check the release notes. You will not receive another alert for 7 days.'
+    : 'This is an optional update. You will not receive another version alert for 7 days.';
   const html = wrapTemplate(
     'AvalancheGo Update Available',
     section(
-      '#3B82F6',
-      'A new optional release is available (recommended but not required)',
+      params.maybeMandatory ? '#F59E0B' : '#3B82F6',
+      heading,
       dataTable(
         dataRow('Validator', name, 'white') +
         dataRow('Running Version', params.currentVersion, '#F59E0B') +
         dataRow('Latest Version', params.latestVersion, '#34D399')
       ) +
       `<p style="font-size: 13px; margin: 12px 0 0 0;">Release notes: ${releaseLink(params.latestVersion)} | Explorer: ${explorerLink(params.nodeId)}</p>`,
-      'This is an optional update. You will not receive another version alert for 7 days.'
+      footnote
     ),
-    '#3B82F6',
+    params.maybeMandatory ? '#F59E0B' : '#3B82F6',
     params.alertId
   );
   return { subject, html, text };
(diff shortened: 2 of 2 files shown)


owenwahlgren:
We should extend the service to show L1 validator AVAX balances

ashucoder9:
> We should extend the service to show L1 validator AVAX balances

we should do this in another PR

server/templates/validator-alerts.ts
```diff
@@ -0,0 +1,272 @@
...
+  daysLeft: number;
+  expiryDate: string;
+  urgency: 'notice' | 'urgent';
+}): { subject: string; html: string; text: string } {
+  const name = params.label ? `${params.label} (${params.nodeId})` : params.nodeId;
+  const prefix = params.urgency === 'urgent' ? '⚠️ ' : '';
+  const subject = `${prefix}Stake Expiry Alert: ${params.nodeId} — ${params.daysLeft} day${params.daysLeft === 1 ? '' : 's'} remaining`;
+  const text = `Your validator ${name} stake expires in ${params.daysLeft} day(s) (${params.expiryDate}).`;
+
+  const borderColor = params.urgency === 'urgent' ? '#EF4444' : '#F59E0B';
```
owenwahlgren, line 232: `"if stake is not renewed"` — stake doesn't get "renewed" on Avalanche. Validators either re-register or their validation period ends. This shows up in both the `notice` and `urgent` expiry templates.

Suggest: `"Next alert in {cooldown} if validation period is not extended."`

Or even better, just say what will happen: `"We'll alert you again in {cooldown} if your validator is still expiring."`

server/services/validator-alert-check.ts
```diff
@@ -0,0 +1,413 @@
...
+
+/**
+ * Cooldown periods in hours per alert type.
+ * Distinct types ensure escalation tiers don't suppress each other.
+ */
+export const COOLDOWNS: Record<AlertType, number> = {
+  uptime: 24,
+  version_mandatory: 48,
+  version_mandatory_urgent: 12,
+  version_mandatory_critical: 4,
```
owenwahlgren, line 23: The 72-hour (3-day) cooldown for `expiry` doesn't make sense relative to the user-configurable `expiry_days` threshold.

Example: user sets threshold to 7 days → gets first alert at 7 days → next alert at 4 days (7 - 3) → then silence until `expiry_urgent` kicks in at 1 day. That's only 2 alerts for someone who specifically asked for a 7-day warning.

Consider either:
- Scaling cooldown relative to `expiry_days` (e.g. `Math.max(12, expiry_days * 8)` hours)
- Or just using 24h cooldown — daily reminders are reasonable for something as important as stake expiry

server/templates/validator-alerts.ts
```diff
@@ -0,0 +1,272 @@
...
+        dataRow('Required Version', params.requiredVersion, '#34D399') +
+        dataRow('Upgrade Deadline', deadlineStr, deadlineColor) +
+        acpLine
+      ) +
+      `<p style="font-size: 13px; color: #EF4444; margin: 12px 0 0 0; font-weight: bold;">Failure to upgrade before the deadline may result in your node being benched.</p>` +
+      `<p style="font-size: 13px; margin: 8px 0 0 0;">Release notes: ${releaseLink(params.requiredVersion)} | Explorer: ${explorerLink(params.nodeId)}</p>`,
+      `Next alert in ${cooldownText} if still not upgraded.`
+    ),
+    borderColor,
+    params.alertId
```
owenwahlgren, line 164: `"Failure to upgrade before the deadline may result in your node being benched."`

This line is the same for all urgency tiers (notice, urgent, critical). For the `critical` tier (< 24h to deadline), "may result" is too soft — it **will** get benched. Should escalate the language:

- notice: "Failure to upgrade before the deadline may result in your node being benched."
- urgent: "Your node will be benched if not upgraded before the deadline."
- critical: "Your node will be benched imminently if not upgraded now."

server/templates/validator-alerts.ts
```diff
@@ -0,0 +1,272 @@
...
+  const minutesLeft = Math.max(1, Math.round(params.hoursLeft * 60));
+  const subject = `🚨 Stake Expires Within 1 Hour: ${params.nodeId}`;
+  const text = `CRITICAL: Your validator ${name} stake expires in approximately ${minutesLeft} minutes (${params.expiryDate}). Immediate action required.`;
+  const html = wrapTemplate(
+    '🚨 Stake Expires Within 1 Hour',
+    section(
+      '#EF4444',
+      'Immediate action required — your stake is about to expire',
+      dataTable(
+        dataRow('Validator', name, 'white') +
```
owenwahlgren, line 278: `"This is a one-time critical alert."` — users don't know what "one-time" means in this context. Does the system give up after this? What if they miss it?

Suggest: `"This is a final reminder — no further alerts will be sent for this expiry."`

server/templates/validator-alerts.ts
```diff
@@ -0,0 +1,272 @@
...
+  return { subject, html, text };
+}
+
+// ---------------------------------------------------------------------------
+// AvalancheGo Upgrade Alert — Mandatory (with escalation)
+// ---------------------------------------------------------------------------
+
+export function versionMandatoryTemplate(params: {
+  alertId: string;
+  nodeId: string;
```
owenwahlgren, line 95: `"You will not receive another uptime alert for this validator within 24 hours."`

This is a double-negative framing that reads a bit like a threat. Suggest:

`"We'll check again in 24 hours and alert you if uptime is still below your threshold."`

server/templates/validator-alerts.ts
```diff
@@ -0,0 +1,272 @@
...
+    section(
+      borderColor,
+      'A mandatory network upgrade is required',
+      dataTable(
+        dataRow('Validator', name, 'white') +
+        dataRow('Running Version', params.currentVersion, '#F59E0B') +
+        dataRow('Required Version', params.requiredVersion, '#34D399') +
+        dataRow('Upgrade Deadline', deadlineStr, deadlineColor) +
+        acpLine
+      ) +
```
owenwahlgren, line 158: When `deadline` is null, this renders as `"Upgrade Deadline: TBD — check release notes"`. 

If we can't determine a deadline, should we really be sending a mandatory-tier email with escalation? The user gets "mandatory upgrade required" but the system can't tell them *when*. This will cause unnecessary anxiety.

Two options:
1. Don't escalate to mandatory tier without a parseable deadline — fall back to optional with a note that the release *may* be mandatory
2. At minimum, change the copy to: `"Deadline not yet announced — monitor release notes"`

server/services/validator-alert-check.ts
```diff
@@ -0,0 +1,413 @@
...
+const P2P_API_URL = 'https://52.203.183.9.sslip.io/api/validators';
+const GITHUB_RELEASES_URL = 'https://api.github.com/repos/ava-labs/avalanchego/releases';
+
+/**
+ * Cooldown periods in hours per alert type.
+ * Distinct types ensure escalation tiers don't suppress each other.
+ */
+export const COOLDOWNS: Record<AlertType, number> = {
+  uptime: 24,
+  version_mandatory: 48,
```
owenwahlgren: 48-hour cooldown for mandatory notice tier is very relaxed. If the deadline is 5 days out, the user gets alert at 5 days, then 3 days, then the `urgent` tier kicks in at 72h.

That's only 2 notice-tier emails. Consider 24h here — a daily reminder for a mandatory network upgrade is reasonable and matches what other monitoring systems do.

components/validator-alerts/AlertDashboard.tsx
```diff
@@ -0,0 +1,394 @@
...
+    }
+  }, [status, fetchAlerts, fetchValidators]);
+
+  // Re-fetch alerts after login completes (fixes post-login refresh issue)
+  useLoginCompleteListener(() => {
+    setLoading(true);
+    fetchAlerts();
+    fetchValidators();
+  });
+
```
owenwahlgren, line 102: `toast.success('Validator added', 'You will receive alerts for this validator.')`

This is slightly misleading — there's no immediate check on creation (alerts run on the cron cycle). If the validator is already below threshold, the user won't know until the next tick.

Either:
1. Change to: `"Validator added. Alerts will begin on the next check cycle (~15 min)."`
2. Or trigger an immediate `checkSingleAlert()` on creation (better UX)

PR #4030, opened 2026-03-26
fix(evaluate): add evaluate link to authorized user profile dropdown

federiconardelli7:
- Add "Evaluate Hackathons" link to user dropdown menu                                                                                                        
  - Visible only to users with `devrel` or `judge` role                                                                                      
  - Missing from PR #4016 which added the /evaluate route and dashboard                                                                                           
                                                                                                                                                                  
  ## Test plan                                                                                                                                                    
  - [ ] Log in as user with `devrel` attribute → link appears in dropdown                                                                                         
  - [ ] Log in as user with `judge` attribute → link appears in dropdown                                                                                        
  - [ ] Log in as regular user → link does not appear                                                                                                             
  - [ ] Click link → navigates to /evaluate

### DIFF
--- a/components/login/user-button/UserButton.tsx
+++ b/components/login/user-button/UserButton.tsx
@@ -133,6 +133,13 @@ export function UserButton() {
                 </DropdownMenuItem>
               )
             }
+            {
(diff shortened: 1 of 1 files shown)


owenwahlgren:
note this is a frontend only change and can be worked around

PR #4098, opened 2026-04-13
fix(console): distinguish chosen vs unchosen branch in stepper

federiconardelli7:
## Summary
  - When a wizard has a branch step (e.g., "L1 Node Setup with Docker" vs "Managed Testnet L1 Nodes"), both options turned green identically after the user moved 
  past the step — no visual indication of which path was chosen                                                                                                   
  - Track the selected branch option in sessionStorage so the choice persists across URL-based step navigation                                                    
  - The chosen option now shows green with a checkmark; the unchosen option stays muted/gray                                                                      
  - "Back" navigation now returns to the previously chosen branch option instead of always defaulting to the first one                                            
  - Affects all console wizard flows with branch steps: L1 create, ICTT setup, ICM setup, staking manager setup                                                   
                                                                                                                                                                  
  ## Test plan                                                                                                                                                    
  - [ ] Navigate to `/console/layer-1/create`, go through steps until "Set Up a Node" branch                                                                      
  - [ ] Select "L1 Node Setup with Docker", click Next — Docker option should be green with checkmark, Managed Testnet should be muted                            
  - [ ] Click Back — should return to Docker option (not Managed Testnet)                                                                                         
  - [ ] Switch to Managed Testnet, click Next — Managed should be green, Docker should be muted                                                                   
  - [ ] Verify same behavior on another flow with branches (e.g., `/console/ictt/setup`)   
<img width="896" height="165" alt="image" src="https://github.com/user-attachments/assets/a97eec04-4a76-4353-b9c5-ac772f145b10" />


owenwahlgren (commented):
## CI Fix Pushed

I pushed a commit adding `components/console/**` to both `console-ci.yml` path triggers and `.lintstagedrc.js` pre-commit config. This is why console-ci wasn't triggering — `components/console/step-flow.tsx` wasn't in the watched paths. It should trigger on the next push now.

---

## Code Review — Changes Needed

The approach is solid (sessionStorage-backed branch choices), but three things need updating to match our conventions:

### 1. Side effect in `setState` updater (React anti-pattern)

```tsx
// Current — impure updater (sessionStorage write is a side effect)
setBranchChoices(prev => {
  if (prev[currentStep.key] === selectedBranchOption.key) return prev;
  const next = { ...prev, [currentStep.key]: selectedBranchOption.key };
  try { sessionStorage.setItem(branchStorageKey, JSON.stringify(next)); } catch {}
  return next;
});
```

React may call updaters multiple times in StrictMode/concurrent rendering. Move the `sessionStorage` write outside:

```tsx
useEffect(() => {
  if (selectedBranchOption && currentStep?.type === "branch") {
    const key = currentStep.key;
    setBranchChoices(prev => {
      if (prev[key] === selectedBranchOption.key) return prev;
      return { ...prev, [key]: selectedBranchOption.key };
    });
    try {
      const stored = JSON.parse(sessionStorage.getItem(branchStorageKey) || '{}');
      stored[key] = selectedBranchOption.key;
      sessionStorage.setItem(branchStorageKey, JSON.stringify(stored));
    } catch {}
  }
}, [selectedBranchOption, currentStep, branchStorageKey]);
```

### 2. Update `getStepNavKey` instead of duplicating logic

There's an existing `getStepNavKey` helper at line ~219 that still defaults to `options[0]`. The PR duplicates the branch-choice-aware logic inline in `prevLink` and the Prev button onClick instead of updating the helper.

Fix: make `getStepNavKey` branch-choice-aware:

```tsx
const getStepNavKey = (step: StepDefinition): string => {
  return step.type === "single"
    ? step.key
    : (branchChoices[step.key] || step.options[0].key);
};
```

Then all four callsites (prevLink, Prev button, both Next buttons) get it for free — and the **Next buttons** (lines ~369, ~399) also get fixed. Right now they still use the old `getStepNavKey` and ignore stored choices.

### 3. Minor: Prev button onClick should use the helper too

After fixing `getStepNavKey`, the Prev button simplifies back to:

```tsx
onClick={() => {
  const prevStep = steps[currentIndex - 1];
  onNavigate(getStepNavKey(prevStep));
}}
```

No inline resolution needed.

PR #4131, opened 2026-04-29
add: research proposals form on Builders Hub 

ashucoder9:
adds form for accepting submissions of research proposals grant


public/images/call-for-research-1.jpeg
nicolasarnedo99, line 1: best to store in vercel blob https://vercel.com/ava-labs/builder-hub/stores/blob/store_QizAt5l3bwvomkny/browser

containerman17 (approved):

PR #4200, opened 2026-05-15
feat(console): show ERC-20 token balances on L1 dashboard

federiconardelli7:
## Summary

  Adds an ERC-20 token balances card to the L1 dashboard so users can see what they actually hold on each L1 — not just the native token. The card surfaces tokens bridged via ICTT (e.g.,
  wrapped AVAX or USDC bridged from C-Chain), the L1's own wrapped native token, and any custom token addresses the user pastes in.

  **What you'll see:** between the existing `HeroCard` (chain identity + native balance) and `L1Details` (health/charts/fleet/etc.), a new `Tokens` card lists every ERC-20 with a non-zero
  balance on the selected L1, with a source badge (`Bridged` / `Wrapped native` / `Custom`), a 0-balance toggle, and a 15s auto-refresh.

  ### How it works

  Three-layer token discovery, all reading directly from the L1's own RPC — no dependency on Glacier indexing the L1:

  - **ICTT bridge index** — a new `/api/l1-tokens/discover?blockchainId=...` endpoint pulls the global ICTT transfers feed (via a new shared 36h cache, reused by `/api/ictt-stats`) and
  returns every candidate ERC-20 address involved in routes touching the L1. Works for any L1 with bridge activity, indexed or not.
  - **L1 metadata** — `L1ListItem.wrappedTokenAddress` is always included when present.
  - **User-added** — paste-an-address modal validates the contract is an ERC-20 on this L1 before persisting to `localStorage` (scoped per `chainId + userAddress`).

  Each candidate is validated on-chain via a single `symbol()/name()/decimals()` viem multicall on the L1's RPC (deployless multicall already enabled in `usePublicClientForChain`). Anything
   that fails the probe is silently dropped — that's how we handle the home-vs-remote ambiguity in the ICTT feed without needing to crack which side of the bridge each address represents.
  Metadata is cached per session; only `balanceOf(user)` re-runs on each 15s refresh.

  ### Files

  **New (7):**
  - `lib/ictt/transfers-cache.ts` — single-flight, stale-on-failure 36h cache for the idx6 ICTT transfers feed
  - `app/api/l1-tokens/discover/route.ts` — GET endpoint, validates blockchainId, returns deduped candidate addresses
  - `lib/console/my-l1/token-discovery.ts` — pure merge/dedupe/normalize helpers
  - `lib/console/my-l1/token-storage.ts` — localStorage wrapper with private-mode-Safari-safe writes
  - `hooks/useL1TokenBalances.ts` — discovery + validation + multicall pipeline + 15s refresh
  - `app/console/my-l1/_components/TokenList.tsx` — card UI with source badges, 0-balance toggle, remove-for-custom dropdown
  - `app/console/my-l1/_components/AddTokenDialog.tsx` — paste-address modal with on-chain pre-validation

  **Modified (2):**
  - `app/api/ictt-stats/route.ts` — reads transfers through the shared cache (preserves its own enriched-response cache layered on top)
  - `app/console/my-l1/_components/DashboardBody.tsx` — slots `<TokenList>` between hero and details

  ### Risks & mitigations

  - **idx6.solokhin.com is a third-party single point of failure.** Mitigated by the 36h cache (returns stale on upstream failure), the soft `tryGetIcttTransfers` path used by the new
  endpoint (returns `{ addresses: [], warning }` instead of 500), and the user-added fallback. We observed Cloudflare 521s during planning so this isn't theoretical.
  - **`contractAddress` vs `coinAddress` in the ICTT feed.** Without cracking which side of the bridge each address represents, the discover endpoint returns both. The on-chain `symbol()`
  probe in the hook transparently drops whichever side isn't a live ERC-20 on the L1.
  - **localStorage drift across wallets sharing a browser.** Keys include the wallet's lowercase EVM address so Alice's custom tokens don't leak into Bob's view.

  ## Test plan

  - [ ] `yarn dev` → open `/console/my-l1?chain=43114` on **mainnet C-Chain** with a wallet that holds ERC-20s. Token list renders below `HeroCard`. Symbols, names, balances, and decimals
  match the block explorer. Refreshes every 15s (verify by spending a small amount elsewhere and waiting).
  - [ ] Switch wallet to a **Glacier-unsupported custom L1 with ICTT activity** (e.g. an L1 that has received bridged tokens from C-Chain). Bridged tokens auto-discover and render with the
  `Bridged` badge.
  - [ ] Click **Add Token**, paste a known ERC-20 contract address on the selected L1 → appears immediately with correct metadata. Reload → token still there. Click `…` → **Remove from
  list** → token disappears; reload confirms removal.
  - [ ] Submit a garbage address (`0xdead…dead`) in **Add Token** → clear inline error *"No ERC-20 token found at this address on {L1 name}"*.
  - [ ] Disconnect wallet (or switch to a chain not in the dashboard) → token list shows *"Switch your wallet to this L1 to see token balances."*
  - [ ] Toggle **Show 0-balance tokens** → previously-hidden tokens with zero balance appear; hidden count updates accordingly.
  - [ ] Simulate idx6 outage (block the network request in DevTools) → discovery degrades silently; user-added flow and wrapped-native candidate continue to work.
  - [ ] `/api/ictt-stats` still returns the same shape after the cache refactor — verify by hitting it directly and confirming `transfers`, `overview`, `topRoutes`, `tokenDistribution` look
   unchanged.
  - [ ] `yarn tsc --noEmit` clean.


lib/ictt/transfers-cache.ts
```diff
@@ -0,0 +1,105 @@
...
+  remoteChainBlockchainId: string;
+  remoteChainName: string;
+  direction: string;
+  contractAddress: string;
+  coinAddress: string;
+  transferCount: number;
+  transferCoinsTotal: number;
+}
+
+const ICTT_TRANSFERS_URL = 'https://idx6.solokhin.com/api/global/ictt/transfers';
```
owenwahlgren, line 24: same here, i think we want to move away from solokhin

owenwahlgren (changes requested):

lib/ictt/transfers-cache.ts
ashucoder9: use: https://github.com/ava-labs/builders-hub/blob/master/lib/ictt-clickhouse.ts

ashucoder9:
regarding the risks and mitigations of using idx6: we should continue using the ictt-clickhouse.ts as a replacement for global/ictt/transfers

it may not work in cases where an L1 isnt indexed by us, but this should be mitigated over time. we shouldnt add back solokhin api

ashucoder9:
@federiconardelli7 pls check if the feature works for L1s that are indexed in our Clickhouse. you can check for CX Chain and KiteAI ICTT numbers as we have coverage and they have some ictt transfers.

PR #4231, opened 2026-05-27
feat(hackathon): submission flow, project blacklist & evaluation UX improvements

alejandro99so:
## Summary

- **Blacklist for devrel** — devrel can hide projects from judges (Hide/Restore button, grayed row, sorted last). Hidden projects are filtered server-side so judges never receive them in the API.
- **Submission status for hackers** — event page shows submission progress (0%, in-progress, complete) with a status card and smart CTA button ("Submit / Continue / Edit project") that updates based on current state. Badge also appears in the top nav bar.
- **Submission confirmation screen** — replaces the profile redirect after final submit with a congratulations screen showing the project name, hackathon, and team member list.
- **Confirmation email** — sent once per project when all required fields are completed on final submit, in the event's language (EN/ES). Controlled by `submission_email_sent` flag to avoid duplicates.
- **Required/Optional field indicators** — all submission form fields now show a red `*` (required) or an "Optional / Opcional" badge, in the event's language.
- **Auto-register on submit** — if a hacker submits a project without having registered for the hackathon, they are auto-registered silently.
- **Sponsors section** — fixed background colors and logo invert filter for light/dark mode.

## Migration

```sql
ALTER TABLE "Project"
  ADD COLUMN "is_blacklisted" BOOLEAN NOT NULL DEFAULT false,
  ADD COLUMN "submission_email_sent" BOOLEAN NOT NULL DEFAULT false;

### CODE REVIEW

File: app/api/project/route.ts [hunk 1]
```diff
@@ -5,12 +5,86 @@ import { GetProjectByHackathonAndUser } from '@/server/services/projects';
...
+
+  try {
+    await prisma.registerForm.create({
+      data: {
+        hackathon: { connect: { id: hackathonId } },
+        user: { connect: { email: session.user.email } },
+        name: user?.name ?? session.user.name ?? '',
+        city: user?.country ?? '',
+        telegram_account: user?.telegram_account ?? '',
+        terms_event_conditions: true,
```
nicolasarnedo99: are we auto-approving the teams and conditions for users with this? if so we shouldn't

server/services/submitProject.ts
```diff
@@ -325,7 +378,7 @@ export async function getProject(projectId: string): Promise<Project | null> {
         email: user?.email ?? member.email ?? "",
         telegram_account: user?.telegram_account ?? null,
         image: user?.image ?? null,
-        custom_attributes: user?.custom_attributes ?? [],
+        custom_attributes: [],
```
nicolasarnedo99: why make this change?

server/services/submitProject.ts
```diff
@@ -247,6 +265,41 @@ export async function createProject(
...
+      // This update will silently fail if the migration hasn't run yet (field unknown),
+      // but the catch block ensures the email still sends on first submit.
+      try {
+        await prisma.project.update({
+          where: { id: savedProject.id },
+          data: { submission_email_sent: true } as any,
+          select: { id: true },
+        });
+      } catch {
+        // Field not yet in DB — migration pending, continue anyway
```
nicolasarnedo99: i dont understand this?? once the code goes live the migration will already have executed

server/services/submitProject.ts
```diff
@@ -247,6 +265,41 @@ export async function createProject(
...
+
+  // Send submission confirmation email once, outside the transaction,
+  // when this is a final submit (not a draft save) and all required fields
+  // are filled and the email hasn't been sent yet.
+  if (
+    !isDraft &&
+    isProjectComplete(projectData) &&
+    savedProject.hackaton_id &&
+    projectData.submittedBy &&
+    !(savedProject as any).submission_email_sent
```
nicolasarnedo99: Minor — once the migration runs and prisma generate re-runs, this as any is no longer needed. Cleanup, not a blocker

lib/hackathons/submission-progress.ts
```diff
@@ -0,0 +1,46 @@
...
+  project_name: string;
+  short_description: string;
+  full_description: string | null;
+  tech_stack: string | null;
+  github_repository: string | null;
+  demo_link: string | null;
+  tracks: string[];
+};
+
+const REQUIRED_FIELDS: Array<keyof RawProject> = [
```
nicolasarnedo99: we have now three sources of truth for "what fields are required to submit":
 - REQUIRED_FIELDS here (7 fields)
 - isProjectComplete() in server/services/submitProject.ts (same 7 fields, re-implemented)
 - BaseFormSchema in components/hackathons/project-submission/hooks/useSubmissionFormSecure.ts (marks only project_name + short_description as required, the rest as .optional())

good moment to consolidate this and export REQUIRED_SUBMISSION_FIELDS from the schema file and import here + in submitProject.ts, or replace isProjectComplete with FormSchema.safeParse(...).success.

server/services/submitProject.ts
```diff
@@ -9,6 +9,22 @@ import { ValidationError } from "./hackathons";
...
 import { Project } from "@/types/project";
 import { Prisma, User } from "@prisma/client";
+import { sendSubmissionConfirmationMail } from "./registerForms";
+
+/** Returns true when all required submission fields are filled in. */
+export function isProjectComplete(p: Partial<Project>): boolean {
+  const hasText = (v: unknown) => typeof v === "string" && v.trim().length >= 1;
+  const hasItems = (v: unknown) => Array.isArray(v) ? v.length > 0 : hasText(v);
+  return (
+    hasText(p.project_name) &&
```
nicolasarnedo99: this is where the duplication on the requirements is happening better to have it all aggregated in one env variable

prisma/schema.prisma
```diff
@@ -167,7 +167,9 @@ model Project {
   is_preexisting_idea Boolean             @default(false)
   small_cover_url     String?             @default("")
   tags                String[]            @default([])
-  is_winner           Boolean?            @default(false)
+  is_winner                Boolean?            @default(false)
+  is_blacklisted           Boolean             @default(false)
```
nicolasarnedo99: might be worth adapting a more general name "is_rejected" just to also englobe projects that might not have submitted certain docs

components/hackathons/event-layouts/ModernEventLayout.tsx
```diff
@@ -146,6 +153,29 @@ export default function ModernEventLayout({
           lang={lang}
         />
         {isHackathon && <HostNavButtons hackathonId={id} />}
+        {isHackathon && isAuthenticated && submissionStatus !== "none" && (
+          <a
```
nicolasarnedo99: better to use next/link's <Link> here — every other navigation in this file uses it, and raw <a> triggers a full page reload

components/hackathons/hackathon/SubmitButton.tsx
```diff
@@ -22,11 +23,16 @@ export default function SubmitButton({
   variant = "red",
   label = "Submit project",
   isAuthenticated = false,
+  projectId = null,
 }: SubmitButtonProps) {
   const { status } = useSession();
   const { openLoginModal } = useLoginModalTrigger();
 
-  const href = customSubmissionLink ?? `/events/project-submission?event=${hackathonId}`;
+  const baseHref = customSubmissionLink
```
nicolasarnedo99: baseHref is assigned and then immediately re-assigned to href with no transformation why not assign to href directly?

components/hackathons/project-submission/components/SubmissionStep1.tsx
```diff
@@ -255,6 +267,8 @@ const SubmitStep1: FC<projectProps> = (project) => {
               <FormLabelWithCheck
                 label='Website'
```
nicolasarnedo99: other label on this form goes through t(lang, ...).

components/hackathons/project-submission/components/SubmissionStep1.tsx
```diff
@@ -338,6 +352,8 @@ const SubmitStep1: FC<projectProps> = (project) => {
               <FormLabelWithCheck
                 label='Socials'
```
nicolasarnedo99: other label on this form goes through t(lang, ...).

app/api/project/route.ts
```diff
@@ -5,12 +5,86 @@ import { GetProjectByHackathonAndUser } from '@/server/services/projects';
...
+    // Auto-register the submitter if they haven't registered for the hackathon yet
+    if (body.hackaton_id) {
+      await autoRegisterIfNeeded(session, body.hackaton_id);
+    }
+
     const newProject = await createProject({ ...body, submittedBy: session.user.email });
-  
-    return NextResponse.json({ message: 'project created', project: newProject }, { status: 201 });
+
+    // Fetch members with user info for the success screen
```
nicolasarnedo99: createProject() already returns the saved project, and now this handler does a second round-trip just to fetch members. 

cleaner: have createProject return members in its result so we don't make two queriesper submission on the hot path.

app/api/projects/[id]/reject/route.ts
```diff
@@ -0,0 +1,36 @@
+import { NextRequest, NextResponse } from "next/server";
```
Comment (nicolasarnedo99) [inline app/api/projects/[id]/reject/route.ts:1]: if you change schema to rejected update this endpoint to /rejected/route.ts

components/hackathons/project-submission/components/FormLabelWithCheck.tsx
```diff
@@ -5,28 +5,41 @@ import { FC } from 'react';
 import { FormLabel } from '@/components/ui/form';
 import { Checkbox } from '@/components/ui/checkbox';
 import { cn } from '@/lib/utils';
+import { EventsLang, t } from '@/lib/events/i18n';
 
 interface FormLabelWithCheckProps {
   label: string;
   checked?: boolean;
+  /** When true shows a red asterisk. When false/undefined shows an "Optional" badge. */
+  required?: boolean;
```
nicolasarnedo99: would make this non-optional

if it is required, set to true, if it isnt set to false - otherwise we might add down the line some fields as required in the schema and then forget about this option when setting the formlabel

components/hackathons/project-submission/components/GeneralSecure.tsx
```diff
@@ -28,6 +29,12 @@ export default function GeneralSecureComponent({
 }) {
   const [step, setStep] = useState(1);
   const [progress, setProgress] = useState(0);
+  type SubmittedMember = { name?: string | null; email?: string | null; role?: string };
```
nicolasarnedo99: SubmittedMember re-declares a shape that already exists as Project.members[number] in types/project.ts. 

The reason it's needed is the upstream widening to project?: any in ProjectSubmissionContext.tsx plus the (result as any).project?.members cast on line 223 — together they erase the type, so this file re-narrows it locally. If ProjectMember ever changes (e.g. email becomes non-nullable, role becomes an enum), this file won't get a TS error because it's running on its own private copy + an any escape hatch. 

To fix/prevent i would type saveProject's return as { ...; project?: Project } in the context, then drop both the SubmittedMember declaration and the as any cast here.

app/api/project/route.ts
Andyvargtz: We need explicit terms and conditions of the event + consent sharing (mandatory for T1 events)

nicolasarnedo99 (dismissed):

nicolasarnedo99 (approved):

PR #4269, opened 2026-06-05
fix(console): allow custom devnet faucet amounts

anishnar:
## Summary
- add a devnet faucet amount input with a 1 to 2,000 AVAX whole-number range
- pass the requested amount through the existing devnet faucet endpoint
- enforce the same 1 to 2,000 AVAX guardrails server-side before sending tokens

## Testing
- ./node_modules/.bin/eslint components/toolbox/console/primary-network/DevnetFaucet.tsx
- git diff --check

Note: app/api/devnet-faucet/route.ts is outside the current ESLint config and is ignored by that command when passed explicitly.

### DIFF
--- a/app/api/devnet-faucet/route.ts
+++ b/app/api/devnet-faucet/route.ts
@@ -5,7 +5,9 @@ import { getAuthSession } from '@/lib/auth/authSession';
 
 const DEVNET_RPC_URL = 'https://api.avax-dev.network/ext/bc/C/rpc';
 const DEVNET_CHAIN_ID = 43117;
-const DRIP_AMOUNT = '2';
+const DEFAULT_DRIP_AMOUNT = 2;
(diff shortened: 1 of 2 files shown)


owenwahlgren (approved):

app/api/devnet-faucet/route.ts
```diff
@@ -62,6 +78,14 @@ export async function GET(request: NextRequest): Promise<NextResponse> {
       );
     }
 
+    const dripAmount = parseDripAmount(request.nextUrl.searchParams.get('amount'));
```
ashucoder9, line 81: @anishnar we should add a ratelimit on the claims

anishnar: Well since its an internal tool i doubt anybody is gonna deplete the reserve.

PR #4283, opened 2026-06-10
feat(qa): agent-drivable QA harness for Academy/Console surfaces

owenwahlgren:
## Summary

QA harness for every surface that embeds console tools — born from FDE-50 (Martin's Academy walkthrough findings). Three tiers:

1. **Static (per-PR, in console-ci):** `scripts/check-academy-embeds.mts` verifies every MDX `@/components/toolbox` import resolves. This was previously unguarded — `tsc` ignores MDX, so console refactors silently broke course pages.
2. **Render smoke (Playwright, `e2e/`):** loads all 55 academy pages with embedded tools, asserts hard signals (HTTP errors, page exceptions, error boundaries, zero tool mounts), screenshots everything for the agent tier.
3. **Agent tier (`e2e/AGENT.md`):** loop-able playbook — judges screenshots/partial-mounts, drives flows interactively, files deduped GitHub issues.

## The wallet shim

`e2e/wallet-shim/core-shim.ts` impersonates Core: private-key-backed EIP-1193 provider as `window.avalanche`/`window.ethereum`, EIP-6963 rdns `app.core` (the exact signal WalletSync uses for the Core path). Implements the `coreViem/rpcSchema.ts` contract including `avalanche_sendTransaction` with **local XP signing** via `@avalanche-sdk/client` `xpAccount` — full console flows run headlessly, no extension, no popups. Fixtures seed wagmi storage so the wallet auto-connects. Mainnet tx submission is refused by design.

## Support changes

- Stable mount-detection attributes: `data-console-tool` (Container), `data-console-tool-gate` (CheckRequirements), `data-console-flow` (StepFlow)
- `e2e/qa-manifest.json` generated by `yarn qa:manifest` (55 academy pages, 157 console routes, all imports currently resolve)
- devDeps: `@playwright/test`, `esbuild`

## First sweep results: 58/61 pass

The 3 failures are a **real pre-existing bug**: `TestSend` crashes with a Rules-of-Hooks violation on all three bridge-tokens course pages (issue filed separately). Also surfaced: `TestSend`/`ChangeWeight` aren't wrapped in `withConsoleToolMetadata` (no chrome/gating), and `getToolboxStore()` creates a fresh persisted zustand store per call (no cache).

## Test plan

- [x] `tsc --noEmit`, eslint (toolbox files), check-console-design, check-academy-embeds all green
- [x] Wallet shim suite 5/5 against local dev
- [x] Academy smoke 55/58 surfaces (3 = real TestSend bug)
[...]

PR #4285, opened 2026-06-10
fix(console): Rules-of-Hooks crash in toolbox stores — TestSend dead on 3 Academy pages

### LINKED ISSUE #4284: [academy-qa] TestSend crashes (Rules-of-Hooks violation) on all bridge-tokens course pages
Found by the QA harness first sweep (PR #4283), reproducible on local dev with a connected wallet.

## Affected routes
- /academy/avalanche-l1/native-token-bridge/01-erc20-to-native/07-bridge-tokens
- /academy/avalanche-l1/native-token-bridge/02-native-to-erc20/08-bridge-tokens
- /academy/avalanche-l1/native-token-bridge/03-native-to-native/07-bridge-native-tokens

## Symptom
The embedded `TestSend` tool (`components/toolbox/console/ictt/token-transfer/TestSend.tsx`) crashes during hydration; the ToolboxMdxWrapper ErrorBoundary catches it and renders the "Try Again" fallback. The course exercise is unusable.

Console errors:
```
React has detected a change in the order of Hooks called by <component>.
The final argument passed to useCallback changed size between renders. Previous: [] Incoming: [[object Object], (arg)=>arg]
TypeError: Cannot read properties of undefined (reading 'length') at areHookInputsEqual
```

## Suspected cause
Hook order is unstable across renders in `TestSend`. Two suspects:
1. `getToolboxStore(destinationSelection.blockchainId)()` at TestSend.tsx:85 — `getToolboxStore` (components/toolbox/stores/toolboxStore.ts:26) creates a **new** zustand store (with `persist`) on every call, and the hook it returns is invoked conditionally on a value that changes at runtime (empty string → selected chain id).
2. Related: `getL1ListStore(...)` calls inside `useL1ByChainId` follow the same dynamic-store-hook pattern.

`TestSend` is also not wrapped in `withConsoleToolMetadata` (no Container chrome, no requirements gating) — same for `ChangeWeight` (`components/toolbox/console/permissioned-l1s/ChangeWeight.tsx`). Worth fixing together for console consistency.

## Repro
```
yarn dev
yarn e2e --grep "bridge-tokens"
# or open any affected route with a connected wallet and watch the embed area
```

owenwahlgren:
Fixes #4284.

## Root cause

Two compounding problems in `toolboxStore.ts`:

1. `getToolboxStore(chainId)` created a **brand-new persisted zustand store on every call** — unlike `getL1ListStore`, which uses singletons.
2. `useToolboxStore()` **early-returned before calling the store hook** when no L1 was selected. When the wallet connected while a tool was mounted (selectedL1 appears), the store subscription appeared mid-lifecycle → React hook order shifted → `TypeError ... at areHookInputsEqual` → component dead behind the ErrorBoundary.

This killed `TestSend` ("Try Again" fallback instead of the exercise) on three live Academy pages:
- `/academy/avalanche-l1/native-token-bridge/01-erc20-to-native/07-bridge-tokens`
- `/academy/avalanche-l1/native-token-bridge/02-native-to-erc20/08-bridge-tokens`
- `/academy/avalanche-l1/native-token-bridge/03-native-to-native/07-bridge-native-tokens`

Only unwrapped tools were exposed: tools behind `CheckRequirements` don't mount until *after* the wallet connects, so they never see the transition.

## Fix

- `getToolboxStore` → per-chainId singletons (mirrors `getL1ListStore`)
- `useToolboxStore` → always calls the store hook; sentinel `NO_CHAIN_SELECTED` store during bootstrap (never written to → no localStorage pollution); still returns inert state pre-selection
- Fixed the same latent bug class at three more sites found by audit:
  - `RemoteInspector` — store hook inside a ternary
  - `DemoInspector` — store hook inside `useMemo` (someone had try/caught the crash; now also properly reactive)
  - `useIcmMigrationOnce` — store hook inside a `useEffect` loop → imperative `getState()`
- Wrapped `TestSend` + `ChangeWeight` in `withConsoleToolMetadata` (standard chrome, Report Issue/Edit links, `WalletConnected` gating). `TestSend`'s inline `Container` moved into tool metadata.

## Verification

Probed all four affected Academy pages with the QA harness wallet shim (auto-connected wallet, the exact crash trigger): tool titles render, zero hook errors, zero page exceptions, no "Try Again". Regression-probed two previously-passing pages (deploy-icm-demo, deploy-native-token-remote): unchanged. `tsc --noEmit`, eslint (toolbox, --max-warnings 0), check-console-design, check-render-safety all green.

Found by the QA harness from #4283 on its first sweep — with that PR's suite, these three pages flip from fail to pass.

### DIFF
--- a/components/toolbox/console/icm/network/hooks/useIcmMigrationOnce.ts
+++ b/components/toolbox/console/icm/network/hooks/useIcmMigrationOnce.ts
@@ -30,7 +30,9 @@ export function useIcmMigrationOnce(): MigrationState {
     for (const l1 of l1List as L1ListItem[]) {
       const existing = chains[l1.id];
       try {
-        const slice = getToolboxStore(l1.id)();
+        // Inside an effect: read store state imperatively — calling the
+        // bound store hook here would violate the Rules of Hooks.
+        const slice = getToolboxStore(l1.id).getState();
         const registryFromToolbox = (slice.teleporterRegistryAddress ?? '') as string;
         const demoFromToolbox = (slice.icmReceiverAddress ?? '') as string;
         const registryFromL1 = (l1.wellKnownTeleporterRegistryAddress ?? '') as string;
--- a/components/toolbox/console/icm/network/inspectors/DemoInspector.tsx
+++ b/components/toolbox/console/icm/network/inspectors/DemoInspector.tsx
@@ -1,10 +1,10 @@
 'use client';
 
-import { useEffect, useMemo } from 'react';
+import { useEffect } from 'react';
 import DeployICMDemo from '@/components/toolbox/console/icm/test-connection/DeployICMDemo';
 import { useIcmSetupStore } from '@/components/toolbox/stores/icmSetupStore';
 import { useL1List, useSelectedL1, type L1ListItem } from '@/components/toolbox/stores/l1ListStore';
-import { getToolboxStore } from '@/components/toolbox/stores/toolboxStore';
+import { getToolboxStore, NO_CHAIN_SELECTED } from '@/components/toolbox/stores/toolboxStore';
 import { Note } from '@/components/toolbox/components/Note';
 import type { Address } from '@/components/toolbox/console/icm/network/types';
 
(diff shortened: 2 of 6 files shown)


ryanpinsker:
Heads-up: #4286 was an accidental parallel fix for the same issue — closed it in favor of this one. One suggestion from that work: consider gating `TestSend` with `WalletRequirementsConfigKey.EVMChainBalance` instead of `WalletConnected` — every ICTT sibling tool (DeployTokenHome, DeployERC20TokenRemote, etc.) uses `EVMChainBalance`, and a send tool can't do anything without gas anyway. Keeps the course's requirement UX consistent step-to-step.

Also FYI: TestSend is embedded on two more pages than the issue lists — `erc20-bridge/.../06-transfer-tokens` and `.../04-multihop` — and the erc20-bridge course isn't in `qa-manifest.json` yet, so the sweep can't see those two. Worth adding to the manifest in #4283.

ashucoder9 (approved):

PR #4286, opened 2026-06-10
fix: Rules-of-Hooks crash in TestSend on bridge-tokens course pages

### LINKED ISSUE #4284: [academy-qa] TestSend crashes (Rules-of-Hooks violation) on all bridge-tokens course pages

ryanpinsker:
Fixes #4284

## Root cause

Two compounding defects in `components/toolbox/stores/toolboxStore.ts`:

1. **`useToolboxStore()` violated the Rules of Hooks.** It returned an inert object early while no L1 was selected — zero store hooks on the bootstrap path, N hooks once `selectedL1` resolved. When wallet hydration flipped `selectedL1` undefined→defined, the hook count changed mid-component and React threw the hook-order error from the issue.
2. **`getToolboxStore(chainId)` was never memoized** — it ran zustand's `create(persist(...))` on every call. Render-time callers (TestSend line 85 et al.) subscribed to a brand-new store every render, and imperative callers (`stores/reset.ts`) mutated throwaway instances that live components never observed (reset wiped localStorage but on-screen state stayed stale).

Why only the TestSend pages crashed: every other console tool is wrapped in `withConsoleToolMetadata`, whose `CheckRequirements` gate delays mounting until wallet/chain state is stable — accidentally masking the flip. TestSend and ChangeWeight mounted raw during hydration.

## Fix

- One memoized store per chain id (module-level `Map`), with an inert **non-persisted** store for the bootstrap `''` key so `useToolboxStore` subscribes unconditionally — stable hook count in every render path.
- Wrap `TestSend` and `ChangeWeight` in `withConsoleToolMetadata` (Container chrome + requirements gating), matching every other console tool. TestSend's hand-rolled `Container` props moved into tool metadata.

## Verification

Ran the #4283 QA harness locally against a dev server (harness merged into a throwaway branch, not part of this PR):

- **Red on master code**: `renders /academy/.../01-erc20-to-native/07-bridge-tokens` fails with the hook-order crash
- **Green with this fix**: all three native-token-bridge pages + `06-change-weight-demo` pass
- `tsc --noEmit` clean

Note: TestSend is embedded on **5** course pages (the three in the issue plus `erc20-bridge/.../06-transfer-tokens` and `.../04-multihop`) — all covered by this fix. The erc20-bridge course is currently absent from the harness's `qa-manifest.json`, so those two pages aren't swept (flagged separately).

### DIFF
--- a/components/toolbox/console/icm/network/hooks/useIcmMigrationOnce.ts
+++ b/components/toolbox/console/icm/network/hooks/useIcmMigrationOnce.ts
@@ -30,7 +30,10 @@ export function useIcmMigrationOnce(): MigrationState {
     for (const l1 of l1List as L1ListItem[]) {
       const existing = chains[l1.id];
       try {
-        const slice = getToolboxStore(l1.id)();
+        // Plain state read — calling the store hook here (inside an
+        // effect) was an invalid hook call that threw, and the catch
+        // below silently skipped the migration for every chain.
+        const slice = getToolboxStore(l1.id).getState();
         const registryFromToolbox = (slice.teleporterRegistryAddress ?? '') as string;
         const demoFromToolbox = (slice.icmReceiverAddress ?? '') as string;
         const registryFromL1 = (l1.wellKnownTeleporterRegistryAddress ?? '') as string;
--- a/components/toolbox/console/icm/network/inspectors/DemoInspector.tsx
+++ b/components/toolbox/console/icm/network/inspectors/DemoInspector.tsx
@@ -1,6 +1,6 @@
 'use client';
 
-import { useEffect, useMemo } from 'react';
+import { useEffect } from 'react';
 import DeployICMDemo from '@/components/toolbox/console/icm/test-connection/DeployICMDemo';
 import { useIcmSetupStore } from '@/components/toolbox/stores/icmSetupStore';
 import { useL1List, useSelectedL1, type L1ListItem } from '@/components/toolbox/stores/l1ListStore';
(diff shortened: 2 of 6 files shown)


ryanpinsker:
Pushed a follow-up commit (e82c3d0) fixing three more call sites of the same `getToolboxStore` misuse class, found while root-causing #4284:

| Where | Bug | User impact |
|---|---|---|
| `ictt/bridge/inspectors/RemoteInspector.tsx` | store hook called conditionally in a ternary | same hook-order crash class as #4284 — the bridge Remote step could die when the destination chain flips |
| `icm/network/inspectors/DemoInspector.tsx` | store hook called inside `useMemo`, error swallowed by `catch` | the toolbox→icmSetupStore demo-address mirror silently never worked |
| `icm/network/hooks/useIcmMigrationOnce.ts` | store hook called inside `useEffect`, error swallowed per-chain | the one-time ICM setup migration silently no-opped for every chain |

The two "silent" ones are worth a close look in review: in both cases a `try/catch` had been eating React's invalid-hook-call error, so the features looked implemented but never ran. With the memoized stores from this PR, `getState()` reads now also hit the same instances live components subscribe to.

ryanpinsker:
Closing in favor of #4285 — Owen fixed #4284 in parallel (his PR landed first, same root cause, same six files, equivalent fixes). Two things from this PR worth carrying over to the review of #4285:

1. **Red-proof exists**: I verified with the #4283 harness that the bridge-tokens render spec *fails* on master code and passes with the store fix — so the spec genuinely exercises the bug and guards the regression.
2. Left a small gating suggestion on #4285 re: `WalletConnected` vs `EVMChainBalance`.

PR #4324, opened 2026-07-01
fix: events/edit - timezone handling

arminreiter:
1. Fixes event start/end dates shifting when the editor is opened from a browser in a different timezone than the event.
2. show validation error on event edit page for empty tags
3. Surfaces Zod validation errors (e.g missing tags) to ui on event edit page

### DIFF
--- a/app/api/events/[id]/route.ts
+++ b/app/api/events/[id]/route.ts
@@ -50,6 +50,15 @@ export const PUT = withAuthRole('devrel', async (req: NextRequest, context: any,
       return NextResponse.json(updatedHackathon);
     }
   } catch (error) {
+    const wrappedError = error as Error;
+    if (wrappedError.cause === 'ValidationError') {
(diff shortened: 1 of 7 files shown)


nicolasarnedo99 (changes requested):
two small things (one change and one check)

**change**

Not introduced by this PR, but related to goals #2/#3 while we're here: 

the edit path PUTs to /api/events/[id], whose catch returns Internal Server Error with a 500 for every failure, including the ValidationError from updateHackathon. Worth a follow-up to catch ValidationError and return { error, details } with a 400, mirroring POST /api/events.

**check**

Heads-up for the description/rollout: because the load path now renders in the event's own timezone instead of the browser's, events created under the old flow will display a different (correct) time after this ships and need a one-time re-save (double check speedrun after we merge)

nicolasarnedo99 (approved):

PR #4333, opened 2026-07-10
fix: mini grants project invites, small refactoring

arminreiter:
- fixes the issue, that invited members can accept a project invite, but invite is not confirmed in db
- add minigrants to top nav (instead of 2x retro 9000 grant)
- updated community events image
- small refactoring of few components:
  - single email validation which is reused across builderhub
  - single member status that is reused across builderhub 
  - single verdict status that is reused across builderhub
  - single component to check project membership
  - few more harmonizations
- add escapeHtml to prevent html injection into malis


lib/email.ts
```diff
@@ -0,0 +1,18 @@
...
+ *
+ * It must be a single definition, not two that merely look alike: the invite
+ * endpoint used to validate with `z.string().email()` while its callers gated on
+ * a looser hand-rolled regex. They disagreed on addresses like
+ * `josé@example.com`, and because Zod validates an array atomically, one such
+ * address rejected an entire batch of invitations.
+ *
+ * Import `emailSchema` in Zod schemas; call `isValidEmail` from UI code.
+ */
+export const emailSchema = z.string().email();
```
nicolasarnedo99, line 14: Swapping the hand-rolled regex for z.string().email() changes validation semantics across every form that used it. 

email.test.ts proves client/server agree, but not backward-compat with the old regex — can you confirm no form now rejects a previously-valid address?

server/services/SendInvitationProjectMember.ts
```diff
@@ -45,9 +54,9 @@ export async function sendInvitation(
 
       <div style="background-color: #27272A; border: 1px solid #EF4444; border-radius: 8px; padding: 20px; margin-bottom: 20px;">
         <p style="font-size: 16px; color: #F87171; margin-bottom: 10px;">
-          <strong>${inviterName}</strong> ${body.replace(`${inviterName} `, '')}
+          <strong>${safeInviterName}</strong> ${safeBody.replace(`${safeInviterName} `, '')}
         </p>
-        <p style="font-size: 20px; font-weight: bold; color: #EF4444; margin: 8px 0;">"${headline}"</p>
+        <p style="font-size: 20px; font-weight: bold; color: #EF4444; margin: 8px 0;">"${safeHeadline}"</p>
         <a href="${inviteLink}" target="_blank" style="display: inline-block; margin-top: 20px; padding: 10px 20px; background-color: #EF4444; color: white; text-decoration: none; border-radius: 4px; font-weight: bold;">
```
nicolasarnedo99: Not changed by this PR, but a member of any project can invite victim@… with hackathon_id = x"><a href="https://evil.test">Claim your grant</a><a href=", breaking out of the attribute to inject an arbitrary link into mail from our sending domain. 

Something like this could fix it: href="${escapeHtml(inviteLink)}". Optionally also validate hackathon_id against the DB on the project path.

arminreiter: fixed

lib/email.ts
arminreiter: added more unit tests just to make sure we cover all cases

Andyvargtz (approved):

PR #4349, opened 2026-07-15
fix(quick-l1): accept 0x-prefixed BLS keys on register-node

owenwahlgren:
The managed-nodes service returns `publicKey`/`proofOfPossession` **0x-prefixed**, so the internal register-node route's strict no-prefix hex check rejected every real payload with `400 Invalid publicKey` — quick-l1 deploys completed fine but never appeared on the user's Managed Testnet Nodes page (and never awarded the node badge).

One-line fix: `/^[0-9a-fA-F]+$/` → `/^(0x)?[0-9a-fA-F]+$/` + corrected the contract comment. Rejects `0x` alone, double-prefix, non-hex, empty (behavior-tested).

Companion to today's quick-l1 IP/provisioning fixes; observed live on deploy `n7fy6an4i8kl3421cdoo1pdb`.

### DIFF
--- a/app/api/internal/quick-l1/register-node/route.ts
+++ b/app/api/internal/quick-l1/register-node/route.ts
@@ -36,7 +36,7 @@ import { prisma } from '@/prisma/prisma';
  *     keeps each route's auth + limits coherent.
  */
 
-const ADDRESS_HEX_RE = /^[0-9a-fA-F]+$/;
+const ADDRESS_HEX_RE = /^(0x)?[0-9a-fA-F]+$/;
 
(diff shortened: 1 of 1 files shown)


ashucoder9 (approved):

owenwahlgren:
Superseded: master already carries this exact fix via cae218d4c ('fix(console): accept 0x-prefixed BLS keys for quick L1 node registry') — after resolving the conflicts toward master, this branch's residual diff is zero lines, so merging would only add an empty merge commit. Closing rather than no-op merging; reopen if the merged status matters for tracking.

PR #4367, opened 2026-07-20
fix(design): address design-investigation findings

owenwahlgren:
## Summary

Mechanical fixes from Federico's Landing v2 Design Investigation (2026-07-20). Four of the six fix-before-merge items; the remaining two need decisions/assets (body typeface strategy, Aeonik Fono from comms).

- **One red token** — the 18 legacy `#e84142`/rgba values in global.css retargeted to brand `#E6212F`; class names unchanged so legacy pages keep working
- **Scroll-snap accessibility** — mandatory snap now requires an 800px-tall viewport; 640-800px falls back to `proximity` so an overflowing chapter can never trap content (WCAG 1.4.4 / 1.4.10)
- **Mobile hero descender clip** — the rotating slot reserves the display face's descent (pb/-mb pair); fixes the 4px shave at 390px
- **OG share card** — statement now renders Aeonik Black (TTF converted from the shipped woff2; satori can't read woff2) and says BUILD A NETWORK (Launch is not our verb)
- **Cold-load font flash** — `Aeonik-Black.woff2` preloaded from the root layout

Report: Landing-v2 Design Investigation, Federico Nardelli, 2026-07-20.

cc @federiconardelli7


federiconardelli7 (approved):
Verified against the repo. Pushed one follow-up commit with two mechanical fixes: the color sweep was case-sensitive and missed six uppercase #E84142 values in global.css (docs subnav and navbar active-tab underlines), and the homepage og alt still said Launch while the card now says BUILD. Also confirmed the TTF is genuine Aeonik Black, the preload href matches the @font-face src, and Geist-Medium.ttf is still used by the other og routes. Non-blocking follow-ups worth a look: scroll-snap-stop always still sits in the 640px media block, so the proximity band may still hard-stop at every chapter; the root-layout preload taxes docs/console routes that never use the display face, and /fonts/* has no long-lived cache-control header, which undercuts the preload on repeat visits; the og TTF is the full 701-glyph commercial cut where a 16-glyph subset (~5KB) would do; the academy og route still carries the old-red logomark; and the new pb/-mb pair on the rotating noun may let the incoming word peek ~2px at swap start (worth a visual check; if real, grow the 105% travel offsets to ~113%).
@owenwahlgren

PR #4383, opened 2026-07-23
update: fix navbar dropdown links, and z-index for explorer secondary navbar

ashucoder9:


owenwahlgren:
Reverted the subnav z-index commit (737bcdbcf) on this branch — the same z-45 → z-35 fix now ships in #4387 alongside the wider explorer styling pass, so keeping it here would have double-landed the change. This PR now carries only the navbar link updates and the Fuji "soon" treatment. (cc @ashucoder9 — done per Owen's call.)

martineckardt (approved):

PR #4412, opened 2026-07-28
fix: update renamed ACP-194 and ACP-236 doc links

owenwahlgren:
Two ACP directories were renamed upstream, which 404s every hand-written link to them:

- `194-streaming-asynchronous-execution` to `194-continuous-execution`
- `236-continuous-staking` to `236-auto-renewed-staking`

## What changed

Updates the six references across the nav config, the SAE components, and two docs pages, including the external `github.com/avalanche-foundation/ACPs` links, which are broken for the same reason.

The generated `content/docs/acps/*.mdx` pages and their `meta.json` are built from the upstream directory names by `utils/remote-content/acps.mts`, so those self-correct on the next `build:remote`. These hand-written references do not.

## Verification

Diffed every `/docs/acps/<slug>` reference in the repo against the live upstream directory listing. These two were the only broken ones, and none remain after this change. `tsc --noEmit` passes.

## Notes

Display labels are deliberately unchanged. "Streaming Asynchronous Execution" is still the name used for the feature throughout these docs, and renaming it to match the ACP's new title is a branding decision rather than a broken link. Happy to follow up if you want the labels aligned.

`.gitignore` lists ACP filenames individually, so the renamed files will not be ignored on the next sync and will show up as untracked. Left out of this PR as a separate concern; a glob would fix it.

### DIFF
--- a/components/navigation/docs-nav-config.tsx
+++ b/components/navigation/docs-nav-config.tsx
@@ -170,13 +170,13 @@ export const acpsOptions: NavOption[] = [
     title: 'Streaming Asynchronous Execution',
     description: 'ACP-194',
     icon: <Book className="w-5 h-5" />,
-    url: '/docs/acps/194-streaming-asynchronous-execution',
+    url: '/docs/acps/194-continuous-execution',
   },
   {
     title: 'Continuous Staking',
     description: 'ACP-236',
     icon: <Book className="w-5 h-5" />,
-    url: '/docs/acps/236-continuous-staking',
+    url: '/docs/acps/236-auto-renewed-staking',
   },
   {
     title: 'ValidatorManager Contract',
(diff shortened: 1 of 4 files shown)


ashucoder9 (approved):

owenwahlgren:
Superseded by the ACP-194 / ACP-236 rebrand already on master. Rebasing this branch onto master produces an empty diff: master fixed all four references, rebranded the display labels (which this PR deliberately left alone), and added redirects in next.config.mjs for both old slugs. Nothing left to merge.

Still stale on master, not covered by this PR: components/explorer-v2/staking/PrimaryStaking.tsx:854 and content/blog/fireblocks-pchain-staking.mdx:281 both point at /docs/acps/236-continuous-staking. They resolve via the redirect, so no 404, but they are not canonical.

PR #4432, opened 2026-07-30
feat(explorer): all-time tick on the shared page clock

owenwahlgren:
## What changed

Adds an ALL segment to the explorer's shared time range control (1D/1W/1M/3M/1Y/ALL) and teaches every stats surface to handle it.

- time-range.tsx: new `all` range (finite 10-year sentinel so window arithmetic stays finite) plus a `rangeWindowLabel` helper so chips read "All time" instead of "Last all time".
- True all-time where the feed supports it: chain metrics (`/api/chain-stats` already pins `all` to the September 2020 genesis timestamp), ICM stats (route's `all` window predates the first ICM message), P-Chain primary metrics, ICM per-chain feeds.
- Labeled clamps where it does not, following the gas page's "longest window / longest computed" rule: overview aggregate (year), token fee history (year), gas history (year/90d), staking money-flow and expiry sheets (year), C-Chain activity classification (90d).
- Small correctness fixes surfaced by the sweep: NetworkIcm's daily average now divides by the days the feed returned rather than the clock's nominal span, and PchainHome's level deltas clamp to the series start so ALL reads "since inception".

## Verification

- tsc --noEmit clean.
- Headless smoke test of all 14 clock-consuming pages with the ALL tick active: zero page errors; every page shows either "All time" (genuine) or a "longest window/computed" clamp label; no "Last all time" anywhere.

## Open questions

- The C-Chain overview's utilization cell clamps to a year under ALL (gas-history API cap) while its strip chip says "All time"; the utilization detail sheet it links to labels its own window. Can add a per-cell note if wanted.

### DIFF
--- a/components/explorer-v2/ExplorerSubnav.tsx
+++ b/components/explorer-v2/ExplorerSubnav.tsx
@@ -187,7 +187,9 @@ function ChainSwitcher({
             />
           )
         )}
-        <span className="max-w-28 truncate font-mono text-[11px] font-bold uppercase tracking-[0.16em] text-zinc-900 sm:max-w-40 md:max-w-56 dark:text-zinc-100">
+        {/* below sm the name would starve the section tabs — the mark and
+            chevron carry the switcher, the page header names the surface */}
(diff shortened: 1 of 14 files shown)


owenwahlgren:
Folded into #4431 so the explorer changes land as one review unit.

PR #4442, opened 2026-08-03
P-Chain explorer: page through an address's full tx history

owenwahlgren:
## What changed
- Address page transactions now page indefinitely: Load more fetches the next 50 rows with the endpoint's `before` block-height cursor and appends, instead of refetching a bigger window capped at 200
- The button actually shows up now: the upstream `truncated` flag is always false (returns false even when `limit` clips the window), so "more exists" is inferred from page fullness
- Cursor overlaps the boundary block by one height and dedupes by hash, so same-block sibling txs are neither skipped nor duplicated

## Verification
- `tsc --noEmit` clean
- Simulated the exact component flow against the live mainnet API on a busy address (`avax19zfygxaf59stehzedhxjesads0p5jdvfeedal0`): three pages, 49 fresh rows per page, one boundary dupe deduped, block ordering monotonic

## Open questions
- The global /txs list and the node page's staking history are separate: /txs ignores `before` upstream (its Load more grows `limit`, which works but refetches from the tip), and the node document caps history at 100 with no params. Both need upstream (Ash API) changes for true cursors.

Stacked on #4438.

### DIFF
--- a/components/explorer-v2/pchain/PchainAddress.tsx
+++ b/components/explorer-v2/pchain/PchainAddress.tsx
@@ -1,6 +1,6 @@
 "use client";
 
-import { useMemo, useState } from "react";
+import { useEffect, useMemo, useState } from "react";
 import Link from "next/link";
 import { ExplorerShell } from "@/components/explorer-v2/ExplorerShell";
 import {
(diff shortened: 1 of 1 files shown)


ashucoder9:
fixed in the feature/explorer-preview branch for address, node, and txs pages

PR #4469, opened 2026-08-11
feat(academy): move quiz progress to the database, resync on login

### LINKED ISSUE #4357: The “Avalanche Fundamentals” course is showing up as “not completed” for me.
It appears that the information on this page might be outdated. Please review and update as needed.

Page: [/academy/avalanche-l1/avalanche-fundamentals/get-certificate](https://build.avax.network/academy/avalanche-l1/avalanche-fundamentals/get-certificate)

On July 1, I completed the first stage of Avalanche Fundamentals at Avalanche Academy (see attached screenshot of the certificate), but today, when I logged in to continue the next stage of the course, I noticed that Avalanche Fundamentals doesn't have the “Completed” icon on the academy's main roadmap. I checked the quiz for the certificate, and two questions in the “Multi-Signature Schemes” section have been automatically unchecked.

_Edited:_
As an additional note, I have my account set up on both Brave and Google Chrome. In Chrome, it was marked as “complete,” but in Brave it wasn't (even though it was the same account). But now, Chrome is showing “not complete” just like Brave. 

Syncing issues?

Image

Bekirerdem:
Quiz progress currently lives only in localStorage/IndexedDB, so it dies with the browser profile — and the Academy certificate flow can lock users out of badges they earned on another device (#4357).

This PR moves quiz progress to the database and keeps the client in sync:

- **`QuizResponse` model + `quizProgress` service** — per-user progress rows, correctness recomputed server-side (the client never gets to claim a right answer).
- **API** — `GET` all progress, `PUT` one response, bulk `POST` for first-visit backfill from the browser's existing local state.
- **Client sync layer** — pulls account rows on load, pushes saves, backfills local history once; signed-out users keep working locally and stop pushing to the API.
- **Badges** — course completion is verified against the DB before an academy badge is awarded.
- **Hardening** — Zod shape validation, rate-limited writes, atomic correct-row freeze, bounded selection arrays.
- **Login resync** (the two issues found in testing, one root cause): the login modal closes in place without a page load, so the sync layer's page-load cache survived login — answers stayed invisible and saves were skipped until refresh. `resyncQuizProgress()` drops both caches and re-runs the sync on the login-complete event and on any unauthenticated→authenticated session transition; the IndexedDB readers re-read once the pull lands. Verified against the reported flows on a live DB; 5 unit tests cover the resync layer (516 green total).

Closes #4357

### DIFF
--- a/utils/quizzes/indexedDB.ts
+++ b/utils/quizzes/indexedDB.ts
@@ -138,6 +138,28 @@ function ensureSynced(): Promise<void> {
   return syncPromise
 }
 
+let resyncInFlight: Promise<void> | null = null
+
+/**
+ * Re-run the account sync after the auth session changes without a page load
+ * (the login modal closes in place — no navigation, so the module state above
+ * survives). Discards the cached one-shot sync AND the cached signed-out
+ * verdict, so the pull sees the account's rows and later saves push again.
+ * Concurrent callers (every mounted quiz fires on the same login event)
+ * share one pass.
+ */
+export function resyncQuizProgress(): Promise<void> {
+  if (typeof window === "undefined") return Promise.resolve()
+  if (!resyncInFlight) {
+    syncPromise = null
+    serverHasAccount = null
+    resyncInFlight = ensureSynced().finally(() => {
+      resyncInFlight = null
+    })
+  }
+  return resyncInFlight
+}
+
 async function pushResponse(quizId: string, response: QuizResponseValue): Promise<void> {
   if (serverHasAccount === false) return // signed out — local-only, no 401 noise
   // Awaited (not fire-and-forget) so that when quiz.tsx awaits
--- a/utils/quizzes/indexedDB.ts
+++ b/utils/quizzes/indexedDB.ts
@@ -45,15 +45,121 @@ function getDBPromise() {
   return dbPromise
 }
 
-export async function saveQuizResponse(quizId: string, response: QuizDB["quizResponses"]["value"]) {
-  if (typeof window !== "undefined") {
-    const db = await getDBPromise()
-    await db.put(quizStoreName, response, quizId)
+type QuizResponseValue = QuizDB["quizResponses"]["value"]
+
+// ---------------------------------------------------------------------------
+// Server sync (issue #4357)
+//
+// IndexedDB stays the local cache and the synchronous source of truth for the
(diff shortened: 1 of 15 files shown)


utils/quizzes/indexedDB.ts
```diff
@@ -45,15 +45,151 @@ function getDBPromise() {
...
+        for (const [quizId, remote] of server) {
+          const mine = local.get(quizId)
+          if (!mine || betterOf(remote, mine) === remote) {
+            const { quizId: _, ...value } = remote
+            await db.put(quizStoreName, value, quizId)
+          }
+        }
+
+        // Local → server: backfill rows the server lacks or trails on
+        const toPush = [...local.entries()]
```
alejandro99so, line 148: ### 🔴 Blocker: cross-account contamination on a shared browser

The IndexedDB store is keyed by `quizId` only (no `user.id`), and the sync layer handles `unauthenticated → authenticated` but never logout or a user switch. So this backfill (lines 118 to 132) pushes the device's local rows to whatever account is currently signed in, without checking they belong to that account.

**Repro:** A completes quizzes on this browser, then A signs out and B signs in. On the next load, `ensureSynced` reads the still-present local rows and POSTs them to B's account (and `autoAwardCompletedCourses` can award B badges A earned). `serverHasAccount` doesn't help, because `resyncQuizProgress` resets it to `null` on session change.

Bekirerdem: fixed in 32dc128. the GET now returns the session's userId and the client records it in a syncMeta store after each signed-in sync — when a different account syncs, the local quiz rows are cleared before the merge/backfill, so nothing crosses the account boundary (the previous owner's copy is already server-side). with the backfill blocked, autoAwardCompletedCourses only ever sees the account's own rows. anonymous progress is still adopted on the device's first sign-in, and a same-account re-login still heals unsent rows. added a regression test for the exact repro plus both boundary cases.

two known limits, both deliberate:

- devices contaminated before this deploy have no recorded owner, so the first account to sign in after the deploy claims the existing local rows once. the alternative was wiping every local store in the migration, which would destroy signed-out users' progress — they have no server copy.
- if a stale tab still shows account A while the browser's cookies have switched to B, an answer given in that tab lands on B. that's cookie-session semantics shared by any web app, not the sync layer — and it only ever carries the single answer the person just gave, never bulk rows.

PR #4494, opened 2026-08-21
fix: access restriction course badge visibility in learning tree

arminreiter:
fix: access restriction course badge visibility in learning tree


components/quizzes/data/index.ts
```diff
@@ -69,6 +69,16 @@ for (const [courseId, data] of Object.entries(courseFiles)) {
   Object.assign(quizzes, data.quizzes);
 }
 
+// The Access Restriction course was split into two certificate halves, but the
+// learning tree still tracks the whole course under its original slug.
+courses['access-restriction'] = {
```
alejandro99so: Could this live in `useCourseCompletion` instead? It's the only consumer, and this adds a course that doesn't exist to shared data, one side effect: `certificates.tsx:135` uses `quizData.courses[courseId]` as the "Course Not Found" gate, so `access-restriction` now passes it, nothing breaks today since the MDX files use the split slugs, but it leaves room for a certificate page on a course that intentionally has no `certificateTemplate`.

components/quizzes/data/index.ts
```diff
@@ -69,6 +69,16 @@ for (const [courseId, data] of Object.entries(courseFiles)) {
...
+  title: 'Access Restriction',
+  quizzes: [
+    ...courses['access-restriction-fundamentals'].quizzes,
+    ...courses['access-restriction-advanced'].quizzes,
```
alejandro99so: Unguarded, and at module top level, if either JSON gets renamed this breaks the `quizData` import across the whole client bundle rather than just the learning tree, `?.quizzes ?? []` would change to "not completed" instead.

hooks/useCourseBadges.ts
```diff
@@ -4,6 +4,12 @@ import { useState, useEffect } from 'react';
 import { useSession } from 'next-auth/react';
 import type { CourseCompletionEntry } from './useCourseCompletion';
 
+// Split courses: the learning tree tracks the whole course under its original
+// slug, but badge requirements key on the certificate halves.
+const badgeCourseAlias: Record<string, string> = {
+  'access-restriction': 'access-restriction-fundamentals',
```
alejandro99so: This alias and the one in `data/index.ts:74` mean different things, that one requires both halves, this one resolves to `-fundamentals` only, it works because badge `1avalancheL1Academy-9access-restriction` lists both requirements, but a fundamentals-only badge maybe would silently pick wrong.

hooks/useCourseBadges.ts
```diff
@@ -42,12 +48,16 @@ export function useCourseBadges(
...
             if (!response.ok) return;
             const data = await response.json();
-            const imagePath = Array.isArray(data)
-              ? data[0]?.image_path
-              : data?.image_path;
+            // A course id can match several badges (per-course badge + the
+            // graduate badge) — the per-course badge has the fewest requirements.
+            const badges = Array.isArray(data) ? data : [data];
+            const imagePath = badges.sort(
+              (a, b) => (a?.requirements?.length ?? 99) - (b?.requirements?.length ?? 99)
```
alejandro99so: This fixes all 9 courses, not just Access Restriction, the `data[0]` it replaces was picking from a `findMany` with no `orderBy` (`server/services/badge.ts:188`), and every course matches two badges, its own and the Graduate, which lists all 10 course_ids, so which image showed up on a node was never actually pinned down.

**Suggestion:** move it into `getBadgeByCourseId` so `assignBadgeToUser` benefits too, and pick by something firmer than `requirements.length`, e.g. the badge whose requirements are a subset of the course's slugs.

alejandro99so (changes requested):

components/quizzes/data/index.ts
arminreiter: true, fixed

hooks/useCourseBadges.ts
arminreiter: implemented the subset-match criterion, but client-side only. `getBadgeByCourseId` keeps returning all matches because `assignBadgeAcademy` iterates them to track graduate badge progress. narrowing it to one badge would break graduate progression.

alejandro99so (approved):

PR #4497, opened 2026-08-21
Fix/various profile issues

arminreiter:
fixes 3 issues with profile/signup:

1. LinkedIn profiles can have special characters (e.g. öüä). This was not allowed and resulted in an error => fixed by f176f963b385fc5504fd810db589d39a08daf145

2. When X account connect failed (e.g. because account is connected to another profile), the UI did not show any issue and X profile was just empty => fixed by 4396fb7aba2edb3cefcbdef874d92d044b40c966 (show error notification)

3. When user use a referral link on mobile, the login dialog and the cookie consent are both visible and cookie consent is not clickable => fixed by 17c79861a7d9c8f6a5bd8e4c5f4cef2874442d76


components/profile/components/hooks/useProfileForm.ts
```diff
@@ -180,11 +181,35 @@ export function useProfileForm() {
...
+          { duration: 10000 },
+        );
+      } else {
+        sonnerToast.error(`Could not connect ${network}. Please try again.`);
+      }
+    };
+    // Defer past this commit's effect phase: sonner's <Toaster> (mounted after
+    // {children} in the root layout) subscribes in its own mount effect, which
+    // runs AFTER this one — a toast dispatched synchronously here is dropped.
+    setTimeout(() => {
```
alejandro99so, line 209: nit: this fires twice in dev, with `reactStrictMode: true` (next.config.mjs:7) and `[]` deps, React 19 runs an extra setup+cleanup cycle on mount, so this schedules two timers and you get two identical toasts on `/profile?x=already_linked`.

Dev-only (production builds don't double-invoke), and sonner already handles it: `create()` dedupes when an id is passed, updating the existing toast instead of stacking a new one, so no restructuring needed, just add an options arg to the three `sonnerToast` calls inside `notify`, keeping the message as-is and leaving the `setTimeout` untouched:

- `sonnerToast.success(msg, { id })`
- `sonnerToast.error(msg, { id })`
- `sonnerToast.error(msg, { id, duration: 10000 })` for the `already_linked` one

with `const id = "link-${network}"` at the top of `notify`, per-network so the X and GitHub toasts don't overwrite each other. It has to be on all three branches, otherwise the success toast still doubles.

Good background on the double-invoke:
https://dev.to/pockit_tools/why-is-useeffect-running-twice-the-complete-guide-to-react-19-strict-mode-and-effect-cleanup-1n60

arminreiter: fixed, thanks!

alejandro99so (approved):

Issue #3973, opened 2026-03-17
sec: CORS wildcard allows cross-origin requests to authenticated endpoints


devrel-ditto:
## Summary

The site serves `access-control-allow-origin: *` on all responses. Several endpoints use cookie-based session auth and handle sensitive operations (faucet transactions, file uploads, admin actions, AI inference). The wildcard CORS policy is overly permissive for these authenticated endpoints.

## Impact

The combination of wildcard CORS and cookie-based auth could allow cross-origin requests from untrusted sites to interact with authenticated endpoints on behalf of logged-in users. The exact exploitability depends on browser credential handling with wildcard origins, but the configuration is unnecessarily broad.

## Recommended Fix

1. **Scope CORS by endpoint type:** Public read-only endpoints (docs, stats, validators) can keep `*`; authenticated endpoints should restrict to `https://build.avax.network`
2. **Add `Vary: Origin`** header to prevent cache poisoning

```javascript
// next.config.mjs
async headers() {
  return [
    {
      source: "/api/:path*",
      headers: [
        { key: "Access-Control-Allow-Origin", value: "https://build.avax.network" },
        { key: "Vary", value: "Origin" },
      ],
    },
  ];
}
```

## Severity

**High** — overly permissive CORS on endpoints handling funds and authenticated operations.

## Related

- #3967 (security headers)
- #3968 (security headers PR)

PR #4510, opened 2026-09-02
fix(proxy): scope CORS to an origin allowlist instead of wildcard

DemhatKara:
## Summary

Addresses the CORS wildcard flagged in #3973. The proxy staged a blanket
`Access-Control-Allow-Origin: *` on its base response; this replaces it with a
small origin allowlist and dynamic reflection.

**What changed**
- New `lib/security/cors.ts` — `getCorsHeaders(origin)` reflects the request
  `Origin` only when it is allowlisted (the production site; `localhost` outside
  production; `*.vercel.app` only on preview deployments), always paired with
  `Vary: Origin`. It never emits `*` and never grants credentials.
- `proxy.ts` — removes the wildcard, applies the scoped headers to the content
  and pass-through responses, and returns proper preflight headers on `OPTIONS`
  (previously a bare `204`) with a 24h `Access-Control-Max-Age`.
- Unit tests for the allowlist boundaries and header logic.

**Scope notes (for transparency)**
- The old `*` was staged on a response object the middleware never actually
  returned, so it was not being served. This is defense-in-depth hardening plus a
  correct, dynamic CORS implementation and a fixed preflight — not a patch for an
  actively exploitable wildcard.
- The middleware matcher does not cover general `/api/*` (faucet, upload, etc.);
  those routes currently emit no `Access-Control-Allow-Origin` at all, so they
  are not cross-origin readable today. I deliberately did not add a blanket
  `next.config.mjs` `/api/*` restriction (as attempted in the now-closed #3975),
  since that would also break legitimately public API/content consumers. Happy to
  layer that on per-endpoint if preferred.
- Credentials are intentionally not granted, so reflected origins can read only
  public, unauthenticated responses.

Refs #3973

### DIFF
--- a/lib/security/cors.ts
+++ b/lib/security/cors.ts
@@ -1,0 +1,94 @@
+/**
+ * Origin allowlisting for the site's Next.js proxy (middleware).
+ *
+ * Replaces a blanket `Access-Control-Allow-Origin: *` that the proxy used to
+ * stage on its response. We reflect the request `Origin` only when it is
+ * trusted, and never grant credentials — so a reflected origin can read only
+ * public, unauthenticated responses. The MCP endpoint keeps its own,
+ * Claude-specific allowlist in `lib/mcp/cors.ts`; this helper governs the proxy.
+ *
+ * See issue #3973.
+ */
+
+// ---------------------------------------------------------------------------
+// Allowed origins
+// ---------------------------------------------------------------------------
+
+/**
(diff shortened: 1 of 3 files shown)


Radeonares32:
Nice

DemhatKara:
Friendly ping — rebased on current master, no conflicts. Happy to adjust the scope (e.g. add the per-endpoint /api/* restriction on top) if you'd prefer that approach.

